PR #28, opened 2017-01-25
Display country package version

### LINKED ISSUE #26: Display the source version of the data
As a user,
I can know which version of the source legislation is used to present the variables,
So that I don't waste time looking at obsolete information

cbenz:
Closes #26

### DIFF
--- a/src/components/pages/home.jsx
+++ b/src/components/pages/home.jsx
@@ -1,44 +1,58 @@
 import {Link} from "react-router"
-import React from "react"
+import React, {PropTypes} from "react"
 
 import BreadCrumb from "../breadcrumb"
+import ExternalLink from "../external-link"
 
 
 const HomePage = React.createClass({
+  propTypes: {
+    countryPackageName: PropTypes.string.isRequired,
+    countryPackageVersion: PropTypes.string.isRequired,
+  },
   render() {
+    const {countryPackageName, countryPackageVersion} = this.props
     return (
       <section>
         <BreadCrumb />
         <div className="page-header">
           <h1>Explorateur de la législation</h1>
         </div>
-        <div className="col-lg-4">
-          <div className="thumbnail">
-            <div className="caption">
-              <h4>Variables et formules socio-fiscales</h4>
-              <p>
-                Visualiser et rechercher parmi les variables d'entrée, les formules et leurs dépendences.
-              </p>
-              <p>
-                <a className="btn btn-primary" href="/graph/">Graphe »</a>
-                <span style={{marginLeft: 20}} />
-                <Link className="btn btn-default" to="/variables">Liste »</Link>
-              </p>
+        <div className="row">
+          <div className="col-lg-4">
+            <div className="thumbnail">
+              <div className="caption">
+                <h4>Variables et formules socio-fiscales</h4>
+                <p>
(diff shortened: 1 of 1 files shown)


fpagnoux:
Rendu :
image

src/components/pages/home.jsx
```diff
@@ -1,45 +1,59 @@
 import {Link} from "react-router"
-import React from "react"
+import React, {PropTypes} from "react"
 
 import BreadCrumb from "../breadcrumb"
+import ExternalLink from "../external-link"
```
fpagnoux: Not a React expert, but I don't see this used further in the file. Is it on purpose ?

src/components/pages/home.jsx
```diff
@@ -1,45 +1,59 @@
...
 
 
 const HomePage = React.createClass({
+  propTypes: {
+    countryPackageName: PropTypes.string.isRequired,
```
fpagnoux: Just so that I understand : where will the value of these two properties will come from ? Which part of the code gets it from the openfisca web api ?

src/components/pages/home.jsx
cbenz: Yes, that's right.

src/components/pages/home.jsx
cbenz: Here are the steps:
- On the server-side, the express server downloads the `/api/1/parameters` and `/api/1/variables` endpoints in parallel via `Promise.all`: [see here](https://github.com/openfisca/legislation-explorer/blob/8a8e9660a5d28552309e0d1c15887da354eff82a/src/server/index.js#L65-L84), then [writes the result to `window.__INITIAL_STATE__`](https://github.com/openfisca/legislation-explorer/blob/8a8e9660a5d28552309e0d1c15887da354eff82a/src/server/html-document.jsx#L26-L28).
- On the client side, the initial state is read from `window.__INITIAL_STATE__`: [see here](https://github.com/openfisca/legislation-explorer/blob/8a8e9660a5d28552309e0d1c15887da354eff82a/src/client.jsx#L16) then passed to the router which calls the right page depending on the URL. The router is in charge to pass these "properties" to the component. See routes [here](https://github.com/openfisca/legislation-explorer/blob/8a8e9660a5d28552309e0d1c15887da354eff82a/src/routes.jsx)

I copied this workflow from documentations of React-Router, that's a quite known way to do with single-page applications in React, and YES it's over-complicated (that's one of the reasons I left those technologies...).

fpagnoux: Ok, thanks for the explanation ! 👍

fpagnoux:
For me it's ok, I don't know if @MattiSG had a more specific idea of how to display the country package version informatin.

src/components/pages/home.jsx
MattiSG: Request for explanations in PRs should be a sign that some refactor / comment / documentation should be added: if it's not clear during review, it probably won't be any clearer when reading the code  :wink:

src/components/pages/home.jsx
MattiSG: I don't understand why this had to be added either.

MattiSG:
> more specific idea of how to display the country package version information

Well I realise I had, but we didn't agree on anything, and this is quite an improvement already! We'll see if we need to improve upon the rendering at some later point  :)

PR #34, opened 2017-01-30
Rename revdisp to revenu_disponible in graph page

cbenz:
As a user,
I can go to https://legislation.openfisca.fr/graph/
Without provoking an exception.

`revdisp` is currently the root node of the displayed tree.
It needs to be renamed to `revenu_disponible` to fix the bug.

### DIFF
--- a/graph/main.js
+++ b/graph/main.js
@@ -1,4 +1,4 @@
-var initialVariableName = 'revdisp';
+var initialVariableName = 'revenu_disponible';
 
 var legislationExplorerBaseUrl = 'https://legislation.openfisca.fr';
 // var legislationExplorerBaseUrl = 'http://localhost:2030';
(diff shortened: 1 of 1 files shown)


graph/main.js
```diff
@@ -7,7 +7,7 @@ var apiBaseUrl = 'https://api.openfisca.fr';
 // var apiBaseUrl = 'http://localhost:2000';
 var variables_url = apiBaseUrl + '/api/1/variables';
 
-var important_variables = ["revdisp", "psoc", "impo", "rev_trav", "rev_cap", "pen", "irpp", "iai", "iaidrdi", "ip_net"];
+var important_variables = ["revenu_disponible", "psoc", "impo", "rev_trav", "rev_cap", "pen", "irpp", "iai", "iaidrdi", "ip_net"];
```
fpagnoux: Some other of these variables have been renamed too.

cbenz:
@fpagnoux You're right, that's a hidden dependency. The `/graph` page is full of debt, let's discuss it tomorrow.

MattiSG:
If there is more work than what has already been done to make this work, let's just drop it and close that PR. No one complained yet :wink:

If there is indeed just to merge this PR to fix, let's proceed 🙂 but that's not what I understand from @fpagnoux' comment…

cbenz:
Indeed, other variables have to be renamed (or removed from the code) but that's not blocking for this PR. I think we can merge to quickly bring back the `/graph` page back to life.

fpagnoux:
_Fine to the shirt_

Issue #23, opened 2017-01-24
Filter variables based on their label


MattiSG:
As a reuser,
I can search for a term in the label of variables,
To identify a specific variable

 - - - -

Look for strict inclusion of the looked for substring in the variable name.

Extends #22.

MattiSG:
Needs [accent-folding](http://alistapart.com/article/accent-folding-for-auto-complete) support.

cbenz:
Didn't we agree on doing a naive and strict substring inclusion?

Could accent-folding be a new issue?

MattiSG:
Ok, you're right, it definitely could. What I meant is I'd rather split across delivering #22, and then #23 with accent-folding, over #22+#23, and then accent-folding.

Issue #22, opened 2016-11-23
Filter variables based on their name


edarin:
As the search box with auto-completion in the graph.

-> to a better UX of the legilsation-explorer

MattiSG:
As a reuser,
I can search for a term in the list of variables,
To identify a specific variable

- - - - -

Look for strict inclusion of the looked for substring in the variable name.

PR #35, opened 2017-01-30
Implement search by name in variables page

cbenz:
Implements #22 and #23 in a row (was easier).

### DIFF
--- a/src/components/pages/variables.jsx
+++ b/src/components/pages/variables.jsx
@@ -3,6 +3,7 @@ import React, {PropTypes} from "react"
 import {Link} from "react-router"
 import {all, append, intersperse, isNil, lensPath, map, merge, over, pipe, prepend, reduce, reject, sortBy, toPairs, values} from "ramda"
 import TreeView from 'react-treeview'
+import removeAccents from 'remove-accents'
 
 import * as AppPropTypes from "../../app-prop-types"
 import BreadCrumb from "../breadcrumb"
@@ -17,9 +18,9 @@ function buildVariablesTree(variables, query) {
     if (!query) {
       return true
     }
-    const lowerCaseQuery = query.toLowerCase()
-    return variable.name.toLowerCase().includes(lowerCaseQuery) ||
-      variable.label && variable.label.toLowerCase().includes(lowerCaseQuery)
+    const normalizedQuery = removeAccents(query.toLowerCase())
+    return removeAccents(variable.name.toLowerCase()).includes(normalizedQuery) ||
+      variable.label && removeAccents(variable.label.toLowerCase()).includes(normalizedQuery)
   }
   function listToFilteredTree(variables) {
     return reduce(
--- a/src/components/pages/variables.jsx
+++ b/src/components/pages/variables.jsx
@@ -3,7 +3,6 @@ import React, {PropTypes} from "react"
 import {Link} from "react-router"
 import {all, append, intersperse, isNil, lensPath, map, merge, over, pipe, prepend, reduce, reject, sortBy, toPairs, values} from "ramda"
 import TreeView from 'react-treeview'
-import removeAccents from 'remove-accents'
 
 import * as AppPropTypes from "../../app-prop-types"
 import BreadCrumb from "../breadcrumb"
@@ -18,9 +17,8 @@ function buildVariablesTree(variables, query) {
     if (!query) {
       return true
     }
-    const normalizedQuery = removeAccents(query.toLowerCase())
-    return removeAccents(variable.name.toLowerCase()).includes(normalizedQuery) ||
-      variable.label && removeAccents(variable.label.toLowerCase()).includes(normalizedQuery)
+    const normalizedQuery = query.toLowerCase()
+    return variable.name.toLowerCase().includes(normalizedQuery)
   }
   function listToFilteredTree(variables) {
     return reduce(
--- a/src/components/pages/variables.jsx
+++ b/src/components/pages/variables.jsx
@@ -1,7 +1,7 @@
 import DocumentTitle from "react-document-title"
 import React, {PropTypes} from "react"
 import {Link} from "react-router"
-import {all, append, intersperse, isNil, lensPath, map, merge, over, pipe, prepend, reduce, reject, sortBy, toPairs, values} from "ramda"
+import {sortBy, toPairs} from "ramda"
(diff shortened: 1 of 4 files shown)


src/components/pages/variables.jsx
```diff
@@ -32,11 +59,21 @@ const VariablesPage = React.createClass({
     countryPackageVersion: PropTypes.string.isRequired,
     variables: PropTypes.arrayOf(AppPropTypes.variable).isRequired,
   },
+  getInitialState() {
+    return {
+      labelFilter: "",
+      nameFilter: ""
```
MattiSG: The `filter` suffix evokes a function to me, not a query string. What about `query` or `lookup`?

src/components/search-box.jsx
```diff
@@ -0,0 +1,55 @@
...
+    labelFilter: PropTypes.string,
+    nameFilter: PropTypes.string,
+    onLabelChange: PropTypes.func,
+    onNameChange: PropTypes.func,
+  },
+  onLabelInputChange(event) {
+    this.props.onLabelChange(event.target.value)
+  },
+  onNameInputChange(event) {
+    this.props.onNameChange(event.target.value)
```
MattiSG: I have limited experience with React, but I thought one of its points was to remove that kind of event forwarding 🤔

src/components/search-box.jsx
```diff
@@ -0,0 +1,55 @@
...
+              className="form-control"
+              id="search-box-name"
+              onChange={this.onNameInputChange}
+              placeholder="Nom"
+              value={this.props.nameFilter}
+            />
+          </div>
+          <div className="form-group">
+            <label htmlFor="search-box-label">
+              Label
```
MattiSG: If the UI is in French, this should be translated too.

src/components/search-box.jsx
```diff
@@ -0,0 +1,55 @@
...
+            </label>
+            <input
+              className="form-control"
+              id="search-box-label"
+              onChange={this.onLabelInputChange}
+              placeholder="Label"
+              value={this.props.labelFilter}
```
MattiSG: Why separate label and name search? I don't think the choice is obvious for the user.

src/components/pages/variables.jsx
```diff
@@ -1,28 +1,55 @@
...
-    (memo, variable) => over(
-      lensPath(getVariablePath(variable)),
-      node => node
-        ? over(lensPath(["variables"]), append(variable), node)
-        : {variables: [variable]},
-      memo,
-    ),
-    {},
-    variables
+  const matchFilters = variable => (!nameFilter || variable.name.toLowerCase().includes(nameFilter.toLowerCase())) && (
```
cbenz: OK for that.

Personally I tend to use `function`s when there is a need to declare variables and/or use control flow like `if` or `for` loops, and `=>` functions when there is none of them (this is more a parametrized expression).

But I'm fine with both.

MattiSG: Well, I think there is a actually a need for `if`, and that it was worked around with `! nameFilter ||`, which is perfectly fine in itself but leads here to a slightly long expression  ;)

src/components/pages/variables.jsx
```diff
@@ -1,28 +1,55 @@
...
+    !labelFilter || variable.label && variable.label.toLowerCase().includes(labelFilter.toLowerCase())
```
cbenz: I'm not sure to understand your sentence :)

src/components/pages/variables.jsx
```diff
@@ -1,28 +1,55 @@
...
   )
+  function removeEmptyNodes(node) {
+    if (isNil(node.children)) {
+      return node
+    } else {
```
cbenz: Personally I would have used a 2 ternary operators (well-indented for readability).

Do you like it or prefer a real `if` with early exit?

src/components/pages/variables.jsx
MattiSG: I'm saying [accent-folding](http://alistapart.com/article/accent-folding-for-auto-complete) should be part of #23 :wink:

src/components/pages/variables.jsx
cbenz: Oops, sorry this does not apply in this case due to the variable declaration in the `else`.

src/components/pages/variables.jsx
```diff
@@ -1,28 +1,55 @@
...
+          ? node
+            ? over(lensPath(["variables"]), append(variable), node)
+            : {variables: [variable]}
+          : node,
+        memo,
+      ),
+      {}
+    ),
+    removeEmptyNodes,
+  )(variables)
```
cbenz: > Why is it so complicated to display a list in React?

This is not related to React.

This function filters the tree leaves, keeping the intermediary non-empty nodes.
It works on data, not the DOM.

Then this data is passed to the DOM in the [`render` method](https://github.com/openfisca/legislation-explorer/pull/35/files/4977ca339fe53affb2f56f32f3c2262d5c8a2f03#diff-d761e4c762b02dbf437c932af71cb9f0R76).

It is expressed using the Ramda library functions (which is like lodash/underscore) and the `pipe` function chains 2 transformations of data. Perhaps this style is complicated but that's a matter of style, and I can rewrite it in a less functional way.

Anyway I'm gonna post a screenshot of the filtered tree to better understand what I did.

src/components/search-box.jsx
cbenz: Hmm that's an interesting discussion (or a troll ^^) but I'm not gonna explain that here.

I'd just say that:
- React is a view library and does not provide any architecture excepted callbacks for events, which are passed to the parent "component"
- Flux architecture has been proposed and its most widespread implementation is the Redux library
- Redux is not used in Legislation Explorer
- `<troll>React+Redux+Webpack+Babel+Ramda+Typescript+... => Elm language + Elm architecture + Elm toolchain</troll>` and I'm now happy with it, and I see through all the comments you do here that you'd like it!

cbenz:
Here is the rendering:

openfisca variables filter

cbenz:
New rendering:

openfisca variables filter 2

src/components/search-box.jsx
fpagnoux: 👍

cbenz:
I removed the label filtering, and renamed this PR as a consequence.

PR #36, opened 2017-01-31
Implement search by label in variables page

### LINKED ISSUE #23: Filter variables based on their label

cbenz:
Fixes #23 

Comes after #35

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "react-intl": "^2.1.2",
     "react-router": "^2.4.1",
     "react-treeview": "^0.4.5",
+    "remove-accents": "^0.4.0",
     "serve-favicon": "^2.3.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "babel-cli": "^6.11.4",
     "bootstrap": "^3.3.5",
     "classnames": "^2.2.5",
+    "diacritics": "^1.3.0",
     "emailjs": "^1.0.5",
     "express": "^4.12.4",
     "intl": "^1.2.4",
@@ -35,7 +36,6 @@
     "react-intl": "^2.1.2",
     "react-router": "^2.4.1",
     "react-treeview": "^0.4.5",
-    "remove-accents": "^0.4.0",
     "serve-favicon": "^2.3.0"
   },
   "devDependencies": {
--- a/src/variables-tree.js
+++ b/src/variables-tree.js
@@ -1,4 +1,5 @@
 import {all, append, intersperse, isNil, lensPath, map, merge, over, pipe, prepend, reduce, reject, values} from "ramda"
+import removeAccents from 'remove-accents'
 
 
 export function buildVariablesTree(variables, query) {
@@ -7,8 +8,9 @@ export function buildVariablesTree(variables, query) {
     if (!query) {
       return true
     }
-    const normalizedQuery = query.toLowerCase()
-    return variable.name.toLowerCase().includes(normalizedQuery)
+    const normalizedQuery = removeAccents(query.toLowerCase())
+    return variable.name.toLowerCase().includes(normalizedQuery) ||
+      variable.label && removeAccents(variable.label.toLowerCase()).includes(normalizedQuery)
   }
   function listToFilteredTree(variables) {
     return reduce(
--- a/src/variables-tree.js
+++ b/src/variables-tree.js
@@ -1,5 +1,5 @@
 import {all, append, intersperse, isNil, lensPath, map, merge, over, pipe, prepend, reduce, reject, values} from "ramda"
-import removeAccents from 'remove-accents'
+import * as diacritics from 'diacritics'
 
 
 export function buildVariablesTree(variables, query) {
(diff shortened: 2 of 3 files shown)


cbenz:
Here is the rendering:

openfisca variables filter 2

As you can see, accent-folding is taken into account.

src/variables-tree.js
```diff
@@ -7,8 +8,9 @@ export function buildVariablesTree(variables, query) {
     if (!query) {
       return true
     }
-    const normalizedQuery = query.toLowerCase()
-    return variable.name.toLowerCase().includes(normalizedQuery)
+    const normalizedQuery = removeAccents(query.toLowerCase())
+    return variable.name.toLowerCase().includes(normalizedQuery) ||
+      variable.label && removeAccents(variable.label.toLowerCase()).includes(normalizedQuery)
```
fpagnoux: This is heavy, and some of the code is duplicated.

cbenz:
I preferred adding a (small) dependency which does the job to remove accents, rather than copy-pasting some dedicated code.

package.json
```diff
@@ -35,6 +35,7 @@
     "react-intl": "^2.1.2",
     "react-router": "^2.4.1",
     "react-treeview": "^0.4.5",
+    "remove-accents": "^0.4.0",
```
fpagnoux: Introducing a new dependency is always a source of complexity. 

I don't really know whether this is a good trade off in our case. Manually taking into account all the cases seems annoying, but the library doesn't seem to be standard either.

@MattiSG what do you think ?

cbenz: [Its implementation](https://github.com/tyxla/remove-accents/blob/master/index.js) seems quite trivial and efficient (uses regexes).

The adherence to this library is small since we use it at a 1 place only with a function with a single parameter.

fpagnoux: image

fpagnoux: image

cbenz:
@fpagnoux @MattiSG What shall we do now about this PR? I don't think we decided yesterday.

package.json
MattiSG: Indeed, [`andrewrk/node-diacritics`](https://github.com/andrewrk/node-diacritics) has [more support](https://github.com/andrewrk/node-diacritics/stargazers), supports [Unicode private range](https://github.com/andrewrk/node-diacritics/blob/master/index.js#L310), and seems more resilient to encoding support, since the charmap is [escaped](https://github.com/andrewrk/node-diacritics/blob/master/index.js#L297). It is also probably more performant, since it doesn't do dynamic RegExes.

I don't think we should do it without a library, though. The charmap won't change much, but there are pesky little issues around it.

MattiSG: [I asked](https://github.com/tyxla/remove-accents/issues/10)  :wink:

MattiSG:
> What shall we do now about this PR?

Keep on working on it. #23 is still on the sprint backlog  :wink:

src/variables-tree.js
```diff
@@ -7,8 +8,19 @@ export function buildVariablesTree(variables, query) {
...
-    return variable.name.toLowerCase().includes(normalizedQuery)
+    function nameMatches(variable, query) {
+      const normalizedName = variable.name.toLowerCase()
+      return normalizedName.includes(query)
+    }
+    function labelMatches(variable, query) {
+      if (!variable.label) {
+        return false
+      }
+      const normalizedLabel = removeAccents(variable.label.toLowerCase())
```
MattiSG: This implementation means that on every search, the whole tree is accent-folded… That's quite a lot of text!
I benched it, it's ok as long as the search is triggered only by an explicit action. Probably won't support search-as-you-type, but currently ok 😉

src/variables-tree.js
```diff
@@ -7,8 +8,19 @@ export function buildVariablesTree(variables, query) {
     if (!query) {
       return true
     }
-    const normalizedQuery = query.toLowerCase()
-    return variable.name.toLowerCase().includes(normalizedQuery)
+    function nameMatches(variable, query) {
+      const normalizedName = variable.name.toLowerCase()
```
MattiSG: Can there be non-lowercase variables? 🤔

fpagnoux: We don't do anything to prevent it, but no, today all variables are lowercase.

package.json
MattiSG: Conclusion of the issue linked above: please switch to `node-diacritics`.

fpagnoux: 👍

cbenz:
Interesting comparison: https://github.com/tyxla/remove-accents/issues/10#issuecomment-276899345

src/variables-tree.js
cbenz: I first tried search-as-you-type and yes, it was slow.

We could store the accent-folded tree, but as it was slow just with name search #35, it seems that it would be premature optimization. The accent folding is not the only responsible.

We could debounce the search-as-you-type too, but let's keep it with an action for now.

src/variables-tree.js
cbenz: Can there be variable name with accents? Same answer ;-)

package.json
cbenz: Right

src/variables-tree.js
MattiSG: No, not the same answer: it is [impossible](https://docs.python.org/2/reference/lexical_analysis.html#identifiers) (as long as the runtime is Python 2) for variable names to contain diacritics. Only a subset of the ASCII range is supported.

cbenz: Oops, I meant "Can there be non-lowercase variable names? Same answer ;-)"

Issue #37, opened 2017-01-31
Add continuous integration


MattiSG:
Run `npm run lint` and `npm run build` on CircleCI.

cbenz:
Service installed on this GitHub repository.

PR #49, opened 2017-02-13
Add CircleCI badge

cbenz:
Connected to #37


MattiSG:
I'm not in favour of adding those kind of badges. They provide very little value to reusers when green (i.e. only “the maintainers use a CI”, which could as well be read from a look at the repo files if you know what it is), but can have a very negative impact if it ever gets red (“wow this package is just broken let's not even try it” even if you don't know what it is).
As for maintainers, there is absolutely no value when compared with notifications.

It's like making your latest health exam public: you should be the first to know if you get sick, but people can only misinterpret the data you're showing.

Issue #38, opened 2017-01-31
Deploy continuously


MattiSG:
Deploy the legislation explorer latest version on every commit to `master`.

Follows #37.

fpagnoux:
#50 has been merged, @cbenz should we close this issue ?

PR #50, opened 2017-02-13
Setup continuous deployment

cbenz:
Connected to #38

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,5 +1,5 @@
 deployment:
   production:
-    branch: continuous-deployment
+    branch: master
     commands:
       - ssh -l openfisca legislation.openfisca.fr "cd legislation-explorer; ./deploy_prod.sh"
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -0,0 +1,5 @@
+deployment:
+  production:
+    branch: continuous-deployment
```
fpagnoux: Should not we deploy on `master` ?

cbenz: Good spot! I was testing and requested reviewing too fast.

fpagnoux: No worries, I know how annoying it is to test continuous deployment :)

cbenz:
@fpagnoux I tested it with the `continuous-deployment` branch, but it should be the same with `master`.

CircleCI offers a very nice experience, with minimal configuration, supporting SSH keys.

PR #56, opened 2017-02-22
Add footer and link to github

### LINKED ISSUE #53: Add a link to GitHub
Add a “Improve this website” link in the footer to foster collaboration.

cbenz:
Fixes #53

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -43,6 +43,12 @@ const App = React.createClass({
               </div>
             </section>
             {this.props.children}
+            <footer style={{marginTop: "5em"}}>
+              <hr/>
+              <a href={config.gitWebpageUrl} rel="external" target="_blank">
+                Améliorer ce site
+              </a>
+            </footer>
           </div>
         </div>
       </DocumentTitle>
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -44,7 +44,7 @@ const App = React.createClass({
             </section>
             {this.props.children}
             <footer>
-              <a href={config.gitWebpageUrl} rel="external" target="_blank">
+              <a href={config.gitWebpageUrl} target="_blank">
                 Améliorer ce site
               </a>
             </footer>
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -44,7 +44,6 @@ const App = React.createClass({
             </section>
             {this.props.children}
             <footer style={{marginTop: "5em"}}>
-              <hr/>
               <a href={config.gitWebpageUrl} rel="external" target="_blank">
                 Améliorer ce site
               </a>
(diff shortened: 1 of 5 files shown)


cbenz:
Renders as:

image

src/components/app.jsx
```diff
@@ -47,6 +47,12 @@ const App = React.createClass({
               </div>
             </section>
             {this.props.children}
+            <footer style={{marginTop: "5em"}}>
+              <hr/>
+              <a href={config.gitWebpageUrl} rel="external" target="_blank">
```
MattiSG: Why add `external`? This will already be inferred by search engines crawlers since the linked URL is absolute and on a separate domain.

src/components/app.jsx
```diff
@@ -47,6 +47,12 @@ const App = React.createClass({
               </div>
             </section>
             {this.props.children}
+            <footer style={{marginTop: "5em"}}>
+              <hr/>
```
MattiSG: Please use CSS to add a `margin` to the `<footer>` over `<hr>`.

src/components/app.jsx
```diff
@@ -47,6 +47,12 @@ const App = React.createClass({
               </div>
             </section>
             {this.props.children}
+            <footer style={{marginTop: "5em"}}>
```
MattiSG: Is it good React practice to do this here? 😱

cbenz: Nothing related to React.

In this project there is no project-wide CSS stylesheet apart from bootstrap.css.

Why 😱? I use inline style when there is no need to generalize a style. If bootstrap 3 had spacing utilities I would have used them, but that's for bootstrap 4 ([spacing](https://v4-alpha.getbootstrap.com/utilities/spacing/))

src/components/app.jsx
cbenz: I thought it was a good practice to put it when it's an external link. OK to remove.

MattiSG: I did not find references to that, and had never seen it. Would you have a source for it to be a good practice? 😃

src/components/app.jsx
MattiSG: Because inline styles are bad for performance, maintainability, and UI unity.

cbenz: I added a stylesheet.

I agree for performance and UI unity (use the same patterns everywhere).
As a side note, about maintainability, I think that putting styles in separate `.css` files is more a separation of technologies, than a separation of concerns. To develop further we could discuss about it IRL not to pollute the PR ;-)

src/components/app.jsx
cbenz: No, I don't have any source and it seems artificial to add it :)

I kept `target="_blank"`.

src/components/app.jsx
MattiSG: I see your point. There's indeed an interesting discussion to have  :)

PR #57, opened 2017-02-22
Change wording of back button and keep search query across navigation

### LINKED ISSUE #54: Ease research iteration
Replace “Back to home” button in the detailed view of a variable with a “Back” button that saves my search.

cbenz:
Fixes #54

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -11,9 +11,9 @@ import {findParametersAndVariables} from "../search"
 
 const App = React.createClass({
   childContextTypes: {
-    query: PropTypes.string,
+    searchQuery: PropTypes.string,
     searchResults: PropTypes.array,
-    setQuery: PropTypes.func,
+    setSearchQuery: PropTypes.func,
   },
   propTypes: {
     children: PropTypes.node.isRequired,
@@ -26,22 +26,22 @@ const App = React.createClass({
   getChildContext() {
     const {parameters, variables} = this.props
     return {
-      query: this.state.query,
+      searchQuery: this.state.searchQuery,
       searchResults: this.state.searchResults,
-      setQuery: query => {
+      setSearchQuery: searchQuery => {
         this.setState({
-          query,
-          searchResults: findParametersAndVariables(parameters, variables, query),
+          searchQuery,
+          searchResults: findParametersAndVariables(parameters, variables, searchQuery),
         })
       },
     }
   },
   getInitialState() {
     const {location, parameters, variables} = this.props
-    const query = location.query.q || ""
+    const searchQuery = location.query.q || ""
     return {
-      query,
-      searchResults: findParametersAndVariables(parameters, variables, query),
+      searchQuery,
+      searchResults: findParametersAndVariables(parameters, variables, searchQuery),
     }
   },
   render() {
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -2,19 +2,47 @@ import url from "url"
 
 import DocumentTitle from "react-document-title"
 import React, {PropTypes} from "react"
+import {locationShape} from "react-router"
 
 import * as AppPropTypes from "../app-prop-types"
 import config from "../config"
+import {findParametersAndVariables} from "../search"
 
 
 const App = React.createClass({
+  childContextTypes: {
+    query: PropTypes.string,
+    searchResults: PropTypes.array,
+    setQuery: PropTypes.func,
+  },
   propTypes: {
     children: PropTypes.node.isRequired,
     countryPackageName: PropTypes.string.isRequired,
     countryPackageVersion: PropTypes.string.isRequired,
+    location: locationShape.isRequired,
     parameters: PropTypes.arrayOf(AppPropTypes.parameterOrScale).isRequired,
     variables: PropTypes.arrayOf(AppPropTypes.variable).isRequired,
-
+  },
+  getChildContext() {
+    const {parameters, variables} = this.props
+    return {
+      query: this.state.query,
+      searchResults: this.state.searchResults,
+      setQuery: query => {
+        this.setState({
+          query,
+          searchResults: findParametersAndVariables(parameters, variables, query),
+        })
+      },
+    }
+  },
+  getInitialState() {
+    const {location, parameters, variables} = this.props
+    const query = location.query.q || ""
+    return {
+      query,
+      searchResults: findParametersAndVariables(parameters, variables, query),
+    }
   },
   render() {
     const {countryPackageName, countryPackageVersion, parameters, variables} = this.props
--- a/src/components/pages/home.jsx
+++ b/src/components/pages/home.jsx
@@ -49,10 +49,11 @@ const HomePage = React.createClass({
     this.context.setSearchQuery(this.state.inputValue)
     this.context.router.push(`?q=${this.state.inputValue}#${searchInputId}`)
   },
-  locationHasChanged(nextLocation) {
-    const {location} = this.props
+  locationHasChanged(location) {
+    const {router} = this.context
+    const oldLocation = this.props.location
     // Check that the new location stays on the Home page, to avoid overwriting searchQuery in App state.
-    if (this._isMounted && location.pathname === nextLocation.pathname) {
+    if (this._isMounted && router.isActive(oldLocation)) {
       const searchQuery = location.query.q || ""
       this.context.setSearchQuery(searchQuery)
       this.setState({inputValue: searchQuery})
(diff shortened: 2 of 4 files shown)


src/components/pages/parameter-or-variable.jsx
```diff
@@ -34,7 +38,7 @@ const ParameterOrVariablePage = React.createClass({
     }
     return (
       <div>
-        <Link className="btn btn-default" to="/">Retour à l'accueil</Link>
+        <Link className="btn btn-default" onClick={this.handleBackButtonClick} to="/">Retour</Link>
```
MattiSG: Why use `onclick` and a handler rather than using `react-router`'s [`goBack()`](https://github.com/ReactTraining/react-router/blob/master/docs/guides/NavigatingOutsideOfComponents.md)?

cbenz: Good catch :) But the solution you linked needs to be called from a `onClick` handler, or what they call a "reducer" but that's for projects using "redux" library, which is not the case here.

So I adapted your proposition, keeping the `onClick` handler, but using the `history` of the router. It's totally identical *in fine*, but perhaps more elegant.

MattiSG: What about simply setting `onClick` to trigger `this.props.history.goBack()`, and changing the `to` to an anchor (`#`), so that you don't have to `preventDefault` and thus to add a wrapper function? Would it work?  🤔

cbenz: The main drawback I see with your proposition is that the user could not copy the link `href`, in an absolute manner (in the sense that going back is relative).

MattiSG: I'm not really sure about this use-case, honestly  :wink: but I see your point…
Then what about interpolating the `href` so that the destination URL is the filled search?  :)  Maybe the `props.history` API exposes a way to get the previous URL?

MattiSG: Well, the current behaviour is worse IMO: on hover, my status bar indicates a misleading destination! You are dynamically hijacking the stated destination.

capture d ecran 2017-02-27 a 10 38 38

The best would be to interpolate the URL directly into the `href` attribute.

MattiSG:
Issue: if I arrive on the details page from a direct link, this “go back” button doesn't mean anything to me.
Solution: “See the `${research results count - 1}` other variables and parameters”.

src/components/pages/parameter-or-variable.jsx
```diff
@@ -34,7 +39,7 @@ const ParameterOrVariablePage = React.createClass({
...
```
MattiSG: Issue: if I arrive on the details page from a direct link, this “go back” button doesn't mean anything to me.
Solution: “See the ${research results count - 1} other variables and parameters”.

MattiSG: Interpolate the URL directly into the `href` attribute so that the browser shows the destination URL.

src/components/pages/home.jsx
```diff
@@ -32,28 +34,32 @@ const HomePage = React.createClass({
...
+    this.context.setQuery(this.state.inputValue)
     this.context.router.push(`?q=${this.state.inputValue}#${searchInputId}`)
   },
-  locationHasChanged(location) {
-    if (this._isMounted) {
-      const query = location.query.q || ""
+  locationHasChanged(nextLocation) {
+    const {location} = this.props
+    // Check that the new location stays on the Home page, to avoid overwriting the query in App state.
+    if (this._isMounted && location.pathname === nextLocation.pathname) {
```
MattiSG: Why not use [`isActive`](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#isactivepathorloc-indexonly)?

src/components/pages/parameter-or-variable.jsx
```diff
@@ -34,7 +39,16 @@ const ParameterOrVariablePage = React.createClass({
     }
     return (
       <div>
-        <Link className="btn btn-default" to="/">Retour à l'accueil</Link>
+        <Link className="btn btn-default" to={"/" + (query ? `?q=${query}#${searchInputId}` : "")}>
```
MattiSG: Why not use [`createPath`](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#createpathpathorloc-query)?

src/components/pages/home.jsx
```diff
@@ -32,28 +34,32 @@ const HomePage = React.createClass({
     return {inputValue: ""}
   },
   handleClearSearchClicked() {
-    this.setState({inputValue: ""})
-    this.context.router.push("")
+    const query = ""
+    this.setState({inputValue: query})
+    this.context.setQuery(query)
+    this.context.router.push(query)
```
MattiSG: This should probably be refactored in a call that handles both clearing and setting the input to avoid duplication and inconsistent behaviour (which is already the case with `scrollIntoView` only called on input change, not on clearing).

src/components/pages/home.jsx
```diff
@@ -32,28 +34,32 @@ const HomePage = React.createClass({
...
   },
   handleInputChange(event) {
     this.setState({inputValue: event.target.value})
     this.searchInput.scrollIntoView()
   },
   handleSubmit(event) {
     event.preventDefault()
+    this.context.setQuery(this.state.inputValue)
     this.context.router.push(`?q=${this.state.inputValue}#${searchInputId}`)
```
MattiSG: Why build a query string when you can use a much more readable [location descriptor](https://github.com/ReactTraining/react-router/blob/master/docs/Glossary.md#locationdescriptor)?

src/components/pages/parameter-or-variable.jsx
cbenz: Well, good remark! I'd like to add a configuration variable, but that would mean updating all the existing links (`<Link to=>` ). Should I create a new issue?

src/components/pages/home.jsx
cbenz: I don't understand if your remark is about line 48 or 49.

On line 48 there is no query string: `this.context.setQuery` sets the *search query* in the global state, accessed through `context` (a React specificity...).

src/components/pages/home.jsx
cbenz: I searched for that and didn't find it 😭  I'll use it

src/components/pages/parameter-or-variable.jsx
MattiSG: Nah, don't do anything if this means adding a configuration variable. I was more thinking of using relative paths like `..` if the relative path can be reliably computed. Defining a root path as a config variable always ends up being a PITA.

src/components/pages/home.jsx
MattiSG: Line 49.

src/components/pages/home.jsx
cbenz: I'm not sure on how to do it. There are many different variations 🤔 

Please come and see me if you'd like to.

I added a comment about `scrollIntoView` in my next commits

src/components/pages/parameter-or-variable.jsx
```diff
@@ -34,7 +39,16 @@ const ParameterOrVariablePage = React.createClass({
...
+          {
+            do {
+              const count = searchResults.length - 1
+              count > 1
+                ? `Voir les ${count} autres variables et paramètres`
+                : "Voir l'autre variable ou paramètre"
```
cbenz: I tried a new formulation. Please tell me your perception.

src/components/app.jsx
```diff
@@ -2,19 +2,47 @@ import url from "url"
...
+import {locationShape} from "react-router"
 
 import * as AppPropTypes from "../app-prop-types"
 import config from "../config"
+import {findParametersAndVariables} from "../search"
 
 
 const App = React.createClass({
+  childContextTypes: {
+    searchQuery: PropTypes.string,
```
MattiSG: > I'm not really fond of the rename in 64040c7a10919fefd40d976c1d393be0441326af, I see longer names with no gain (actually, `searchQuery` sounds redundant to me, and less clear than `query`).
> Anyway, that's not really important.

MattiSG:
Please don't force-push while in the review process. This makes it impossible to track changes since last review, thus making the reviewer's job very hard.

cbenz:
@MattiSG I tested by replacing [this line](https://github.com/openfisca/legislation-explorer/blob/244632fc70adba07ae250471bcd89dc303e67adb/src/search.js#L23) by:

```diff
-  return concat(matchesInName, matchesInDescriptionOnly)
+  return [concat(matchesInName, matchesInDescriptionOnly)[0]]
```

Here is the result:

image
----
image

PR #58, opened 2017-02-22
Improve CTA discoverability

### LINKED ISSUE #51: Improve CTA discoverability
As an OpenFisca reuser,
I can understand the expected result of the search button,
So that I save a seconds on my first interactions

- - - -

- The “Trouver” button should be in a contrast color.
- The “En savoir plus” button should be a link.
- The logo should be smaller.

cbenz:
Fixes #51

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -27,7 +27,7 @@ const App = React.createClass({
               <div className="row">
                 <div className="col-md-3">
                   <img
-                    alt="Logo OpenFisca"
+                    alt=""
                     src={url.resolve(config.websiteUrl, "/hotlinks/logo-openfisca.svg")}
                     style={{maxWidth: 200}}
                   />
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -27,7 +27,7 @@ const App = React.createClass({
               <div className="row">
                 <div className="col-md-3">
                   <img
-                    alt=""
+                    alt="OpenFisca"
                     src={url.resolve(config.websiteUrl, "/hotlinks/logo-openfisca.svg")}
                     style={{maxWidth: "12em"}}
                   />
@@ -35,8 +35,8 @@ const App = React.createClass({
                 </div>
                 <div className="col-md-9">
                   <p>
-                    OpenFisca référence {variables.length} variables et {parameters.length} paramètres
-                    qui modélisent le système socio-fiscal français.
+                    {variables.length} variables et {parameters.length} paramètres
+                    référencées pour modéliser le système socio-fiscal français.
                   </p>
                   <a href={config.websiteUrl}>En savoir plus</a>
                 </div>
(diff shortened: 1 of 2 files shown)


cbenz:
Renders as:

image

src/components/app.jsx
```diff
@@ -25,24 +25,20 @@ const App = React.createClass({
...
-                  <p>
-                    <img
-                      alt="Logo OpenFisca"
-                      className="img-responsive"
-                      src={url.resolve(config.websiteUrl, "/hotlinks/logo-openfisca.svg")}
-                    />
-                  </p>
+                <div className="col-md-3">
+                  <img
+                    alt="Logo OpenFisca"
```
MattiSG: You have to think of the `alt` attribute as a text replacement that will be read to a screenreader, for example. Nobody cares that it's a logo.
As [decorated text](https://www.w3.org/QA/Tips/altAttribute), the `alt` attribute should have only a value of “OpenFisca”, not “Logo OpenFisca”. 
However, in the current case, since you repeat “OpenFisca” in the sentence right after it (which might not be so useful, btw), the image is decorative only and you should use an [empty `alt`](https://www.w3.org/TR/WCAG20-TECHS/H67.html).

src/components/app.jsx
```diff
@@ -25,24 +25,20 @@ const App = React.createClass({
...
               <div className="row">
-                <div className="col-lg-3">
-                  <p>
-                    <img
-                      alt="Logo OpenFisca"
-                      className="img-responsive"
-                      src={url.resolve(config.websiteUrl, "/hotlinks/logo-openfisca.svg")}
-                    />
-                  </p>
+                <div className="col-md-3">
```
MattiSG: Are you sure you need this `<div>` wrapper? What happens if you set the `col` class directly on the ``?

src/components/app.jsx
```diff
@@ -25,24 +25,20 @@ const App = React.createClass({
...
+                  <img
+                    alt="Logo OpenFisca"
+                    src={url.resolve(config.websiteUrl, "/hotlinks/logo-openfisca.svg")}
+                    style={{maxWidth: 200}}
```
MattiSG: Since the layout is mostly based on text, this [should not be set in `px` but in `em`](http://stackoverflow.com/questions/609517/why-em-instead-of-px), to adjust gracefully to different font sizes.

src/components/app.jsx
cbenz: Hmm, there's also `<p>{countryPackageName}@{countryPackageVersion}</p>` in this `<div>`, did you notice? 🤔

src/components/app.jsx
cbenz: > you repeat “OpenFisca” in the sentence right after it (which might not be so useful, btw)

I tried and it's quite strange for the sentence to start by "référence ...".

I propose to keep the logo, use `alt="OpenFisca"` and display the sentence "1468 variables et 1844 paramètres modélisent le système socio-fiscal français.".

src/components/app.jsx
MattiSG: Oops, I had missed it, thanks!  :)

src/components/app.jsx
MattiSG: “modélisent” -> “référencées pour modéliser”, and 👍

PR #46, opened 2017-02-10
Redesign home page

cbenz:
Connected to #40 

Fixes #29

PR #59, opened 2017-02-22
Improve the search algorithm

### LINKED ISSUE #52: Improve search results ordering
The current ordering is only alphabetical.

There should be a sorting algorithm that priories matches according to the position of the match in the variable name (the earlier the better).

cbenz:
Connected to #52

### DIFF
--- a/src/search.js
+++ b/src/search.js
@@ -3,25 +3,26 @@ import * as diacritics from 'diacritics'
 
 
 export function findParametersAndVariables(parameters, variables, query) {
-  function indexInString(substring, string) {
+  function weightIn(string, substring) {
     const index = string.indexOf(substring)
     return index === -1 ? Number.MAX_SAFE_INTEGER : index
   }
   const normalizedQuery = diacritics.remove(query.toLowerCase())
-  const sortFunction = isEmpty(normalizedQuery)
-    ? sortBy(prop('name'))
-    : sortWith([
-      ascend(item => indexInString(normalizedQuery, item.name)),
-      descend(item => item.searchIndex.includes(normalizedQuery)),
-      ascend(prop('name')),
-    ])
-  return sortFunction(parametersAndVariables)
+  const parametersAndVariables = concat(parameters, variables)
+  if (isEmpty(normalizedQuery)) {
+    return sortBy(prop('name'), parametersAndVariables)
+  }
+  return sortWith([
+    ascend(item => weightIn(item.name, normalizedQuery)),
+    descend(item => item.normalizedDescription.includes(normalizedQuery)),
+    ascend(prop('name')),
+  ], parametersAndVariables)
 }
 
-export function withSearchIndex(propertyName, objects) {
+export function addNormalizedDescription(propertyName, objects) {
   return map(
     object => assoc(
-      'searchIndex',
+      'normalizedDescription',
       has(propertyName, object)
         ? diacritics.remove(object[propertyName].toLowerCase())
         : '',
--- a/src/search.js
+++ b/src/search.js
@@ -8,10 +8,6 @@ export function findParametersAndVariables(parameters, variables, query) {
     return index === -1 ? Number.MAX_SAFE_INTEGER : index
   }
   const normalizedQuery = diacritics.remove(query.toLowerCase())
-  let parametersAndVariables = concat(
-    map(assoc('type', 'parameter'), parameters),
-    map(assoc('type', 'variable'), variables),
-  )
   const sortFunction = isEmpty(normalizedQuery)
     ? sortBy(prop('name'))
     : sortWith([
--- a/src/search.js
+++ b/src/search.js
@@ -1,23 +1,25 @@
-import {assoc, concat, filter, has, isEmpty, map, pipe, prop, sortBy} from "ramda"
+import {ascend, assoc, concat, descend, has, isEmpty, map, prop, sortBy, sortWith} from "ramda"
 import * as diacritics from 'diacritics'
 
 
 export function findParametersAndVariables(parameters, variables, query) {
-  function matchesQuery(parameterOrVariable) {
-    if (isEmpty(normalizedQuery)) {
-      return true
-    }
-    return parameterOrVariable.name.includes(normalizedQuery) ||
-      parameterOrVariable.searchIndex.includes(normalizedQuery)
+  function indexInString(substring, string) {
+    const index = string.indexOf(substring)
+    return index === -1 ? Number.MAX_SAFE_INTEGER : index
   }
   const normalizedQuery = diacritics.remove(query.toLowerCase())
-  return pipe(
-    filter(matchesQuery),
-    sortBy(prop('name')),
-  )(concat(
+  let parametersAndVariables = concat(
     map(assoc('type', 'parameter'), parameters),
     map(assoc('type', 'variable'), variables),
-  ))
+  )
+  const sortFunction = isEmpty(normalizedQuery)
+    ? sortBy(prop('name'))
+    : sortWith([
+      ascend(item => indexInString(normalizedQuery, item.name)),
+      descend(item => item.searchIndex.includes(normalizedQuery)),
+      ascend(prop('name')),
+    ])
+  return sortFunction(parametersAndVariables)
 }
 
 export function withSearchIndex(propertyName, objects) {
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -4,7 +4,7 @@ import favicon from "serve-favicon"
 import path from "path"
 
 import handleRender from "./render"
-import {withSearchIndex} from "../search"
+import {addNormalizedDescription} from "../search"
 import {fetchParameters, fetchVariables} from "../webservices"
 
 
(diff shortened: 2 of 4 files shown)


cbenz:
Rendered as;

image

src/search.js
```diff
@@ -1,23 +1,25 @@
...
 
 
 export function findParametersAndVariables(parameters, variables, query) {
-  function matchesQuery(parameterOrVariable) {
-    if (isEmpty(normalizedQuery)) {
-      return true
-    }
-    return parameterOrVariable.name.includes(normalizedQuery) ||
-      parameterOrVariable.searchIndex.includes(normalizedQuery)
+  function indexInString(substring, string) {
```
MattiSG: This function is not properly named: it does _not_ return the index of a substring in a string.

src/search.js
```diff
@@ -1,23 +1,25 @@
...
+    return index === -1 ? Number.MAX_SAFE_INTEGER : index
   }
   const normalizedQuery = diacritics.remove(query.toLowerCase())
-  return pipe(
-    filter(matchesQuery),
-    sortBy(prop('name')),
-  )(concat(
+  let parametersAndVariables = concat(
     map(assoc('type', 'parameter'), parameters),
     map(assoc('type', 'variable'), variables),
```
MattiSG: Why add a `type` on every search? This property does not seem to be used in this function.

src/search.js
```diff
@@ -1,23 +1,25 @@
...
-  ))
+  )
+  const sortFunction = isEmpty(normalizedQuery)
+    ? sortBy(prop('name'))
+    : sortWith([
+      ascend(item => indexInString(normalizedQuery, item.name)),
+      descend(item => item.searchIndex.includes(normalizedQuery)),
```
MattiSG: Where does this `searchIndex` property come from? I guess the parameters of `findParametersAndVariables` have been applied `withSearchIndex`, but this is super unreadable.

Where else is this property needed? If it is used only here, it should not be the responsibility of the caller to add it, but of the search function itself.

src/search.js
```diff
@@ -1,23 +1,25 @@
...
+      ascend(prop('name')),
+    ])
+  return sortFunction(parametersAndVariables)
 }
 
 export function withSearchIndex(propertyName, objects) {
```
MattiSG: This is _really_ hard to understand. Please at least rename the function so that it has an active verb that makes its action clear, and the arguments so that one can get an idea of what type of objects and property names are expected to be passed.
What I get is that you add a `searchIndex` property to each object in an array of objects, which can be either an empty string (then why the hell is it called `searchIndex`? I would expect a Number in there!) or a String taken from some property of that object with removed diacritics. It takes me 5 minutes to decode that algorithm and I don't understand the intention at all  😞

cbenz:
Sorry, I didn't tell but this highlight is just Firefox one, and was intended to help the reviewers to evaluate the algorithm.

I could add a real highlight though.

src/search.js
cbenz: There is a dumb mistake, sorry. I'm gonna fix it but it should be "description" and not searchIndex! 

The odd part is that it works as if it *was* "description"... I'll fix that very soon.

MattiSG:
> this highlight is just Firefox one

Aaah ok, since you wrote “rendered as” I thought this was the actual rendering  :)

MattiSG:
> I could add a real highlight though.

Would be cool, but not in this PR, and definitely not overriding any of the other prioritised items in this sprint 😉

src/search.js
cbenz: OK.

After having re-read the whole thing, it appears that the `searchIndex` property is the description without diacritics. It's not added by the search function because it's bad for performance as the same computation would be done on every search. It's done once at application start, when it fetches the API data from `parameters` and `variables` endpoints.

Perhaps it should be named `normalizedDescription`, but parameters have labels, so `normalizedDescriptionOrLabel` should be more exact, but ugly.

src/search.js
cbenz: Indeed, that should have been deleted in #46.

It was used previously to determine the URLs which included `/variables` or `/parameters` prefix (ie `http://localhost:2030/variables/irpp`), but since we decided to have `http://localhost:2030/irpp` this is no more necessary to add a `type`.

But you're right, even previously, this should have been done in the caller.

cbenz:
I tried to make the search algo more understandable, please tell me if that's better.

src/search.js
```diff
@@ -1,31 +1,31 @@
...
 
 
 export function findParametersAndVariables(parameters, variables, query) {
-  function matchesQuery(parameterOrVariable) {
-    if (isEmpty(normalizedQuery)) {
-      return true
-    }
-    return parameterOrVariable.name.includes(normalizedQuery) ||
-      parameterOrVariable.searchIndex.includes(normalizedQuery)
+  function weightByPositionIn(string, substring) {
```
MattiSG: > Do you really need to specify in the function name _how_ the weight is computed, when there currently is only one way to weigh strings? Maybe `weightIn` is enough here? :)
> Also, this is not exactly the _weight_ in the usual sense, since the higher it is the lower the importance :wink: “inferiority”, maybe? 🤔

src/search.js
```diff
@@ -1,31 +1,31 @@
...
+    return sortWith([
+      ascend(item => weightByPositionIn(item.name, normalizedQuery)),
+      descend(item => item.normalizedDescription.includes(normalizedQuery)),
+      ascend(prop('name')),
+    ], parametersAndVariables)
+  }
 }
 
-export function withSearchIndex(propertyName, objects) {
+export function withNormalizedDescription(objects) {
```
MattiSG: I would recommend to always [use a verb](http://stackoverflow.com/a/1991341) in function names that have side effects. In that case, `withNormalizedDescription` -> `addNormalizedDescription`.

src/search.js
```diff
@@ -1,31 +1,31 @@
...
-    filter(matchesQuery),
-    sortBy(prop('name')),
-  )(concat(
-    map(assoc('type', 'parameter'), parameters),
-    map(assoc('type', 'variable'), variables),
-  ))
+  const parametersAndVariables = concat(parameters, variables)
+  if (isEmpty(normalizedQuery)) {
+    return sortBy(prop('name'), parametersAndVariables)
+  } else {
```
MattiSG: You don't need an `else` when your first `if` contains a `return` 😃
That's an early exit and makes for a much more readable code by avoiding unnecessary indentation and mental conditions.

src/search.js
```diff
@@ -1,31 +1,31 @@
...
+export function withNormalizedDescription(objects) {
   return map(
     object => assoc(
-      'searchIndex',
-      has(propertyName, object)
-        ? diacritics.remove(object[propertyName].toLowerCase())
+      'normalizedDescription',
+      has('description', object)
+        ? diacritics.remove(object.description.toLowerCase())
         : '',
```
MattiSG: > This whole ternary could also just be `diacritics.remove(object.description || '').toLowerCase()`  😉

src/server/index.js
```diff
@@ -73,8 +73,8 @@ Promise.all([fetchParameters(), fetchVariables()])
       countryPackageName: country_package_name,
       countryPackageVersion: country_package_version,
       currency,
-      parameters: withSearchIndex('description', parameters),
-      variables: withSearchIndex('label', variables),
+      parameters: withNormalizedDescription('description', parameters),
+      variables: withNormalizedDescription('label', variables),
```
MattiSG: This call doesn't seem to be legit, since `withNormalizedDescription` only takes a single parameter.

src/search.js
cbenz: I agree. That's the reason I preferred using a ternary operator: to have one code path only :) But the drawback was to have to name the sort function, and `sortFunction` is poor naming.

src/search.js
cbenz: I agree, but in that case the input data is not mutated. In a language which allows both mutating and non-mutating functions, I like to use:
- `addFoo(obj)` when doing `obj.foo = "bar"`
- `withFoo` when doing `Object.extend({}, foo, "bar")` (or `assocFoo` to follow RamdaJS naming, but can be considered odd)

> One of the more universal, yet simple rules is: Function names should be verbs if the function changes the state of the program, and nouns if they're used to return a certain value.

Besides, I think in that case that this function matches the second part of the sentence, but that's a modified copy of the input, too :)

So OK for `addNormalizedDescription` and keep in mind it's not a mutating function!

Sorry for this long answer for a such small point :-D

src/search.js
MattiSG: No, the ternary and the early return are different implementations. Ternaries for long values are unreadable.

What I'm saying is you can write the discussed piece of code as follows:

```js
if (isEmpty(normalizedQuery))
    return sortBy(prop('name'), parametersAndVariables)

// do something that's in the `else` block: it will only be reached when the `if` block was not reached
```

src/search.js
MattiSG: Ok, I'm not used to ramda and thought that `assoc` was mutating the object.

MattiSG: I definitely agree with your point, `add` is not proper here. `cloneWithNormalizedDescription` would satisfy both our points  :wink:

MattiSG:
It seems there was an issue with diff rendering in GitHub, I just commented on stuff you had already fixed 😕

cbenz:
@MattiSG I forget to do `npm install` to upgrade the dependencies expressed in `package.json`.

In this case it's because the `ascend` function was added in a recent version of Ramda.js.

MattiSG:
Should get a bonus for exact match:

capture d ecran 2017-02-28 a 14 56 10

I leave it up to you to see if this is included in this PR or not.

PR #60, opened 2017-03-02
Update dependencies before build in deployment script

cbenz:
This improves resilience of continuous deployment script (introduced in #50), in the case there is a need to update the dependencies.

> I could have added a `npm install` line in [deploy_prod.sh](https://github.com/openfisca/legislation-explorer/blob/master/deploy_prod.sh) before `npm run build`, but the `prebuild` script is less specific.

### DIFF
--- a/deploy_prod.sh
+++ b/deploy_prod.sh
@@ -5,6 +5,7 @@ pwd
 git branch
 
 git pull
+npm install
 npm run build
 
 echo "Restarting legislation-explorer service..."


cbenz:
@MattiSG What about adding a `npm install` line in [deploy_prod.sh](https://github.com/openfisca/legislation-explorer/blob/master/deploy_prod.sh) before `npm run build`, then? This script is only for CircleCI which is not offline.

This PR aims to improve resilience of deploy-to-production.

MattiSG:
Yes, in this case you need to add it to the deployment script. This is not so much of increasing resilience as actually fixing the process, I believe  :wink:

MattiSG:
Merging with “changes requested” is not very nice 😕

cbenz:
I understood in your previous comment that it's OK to add `npm install` line in `deploy_prod.sh` before `npm run build`. That's what I did, did you notice?

MattiSG:
Yes, I did notice, but only after a diff double-check. It is not a good habit to take to merge without a green tick — you never know when the agreement was clear for everybody 🙂 In this case it was, and there was no harm done, but I'd really prefer we don't rely on that :wink:

PR #62, opened 2017-03-02
Improve search results ordering

### LINKED ISSUE #52: Improve search results ordering

cbenz:
Fixes #52

### DIFF
--- a/src/components/pages/home.jsx
+++ b/src/components/pages/home.jsx
@@ -112,15 +112,13 @@ const SearchResults = React.createClass({
     return (
       <List items={items} type="unstyled">
         {parameterOrVariable => {
-          const {name, type} = parameterOrVariable
-          const description = type === 'parameter'
-            ? parameterOrVariable.description
-            : parameterOrVariable.label
+          const {description, itemType, label, name} = parameterOrVariable
+          const displayedDescription = itemType === 'parameter' ? description : label
           return (
-            <Link key={`${name}-${type}`} to={`/${name}`}>
+            <Link key={`${name}-${itemType}`} to={`/${name}`}>
               <article style={{margin: "3em 0"}}>
                 <h4>{name}</h4>
-                {description && <p>{description}</p>}
+                {displayedDescription && <p>{displayedDescription}</p>}
               </article>
             </Link>
           )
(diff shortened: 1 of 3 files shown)


src/components/pages/home.jsx
```diff
@@ -112,15 +112,13 @@ const SearchResults = React.createClass({
...
         {parameterOrVariable => {
-          const {name, type} = parameterOrVariable
-          const description = type === 'parameter'
-            ? parameterOrVariable.description
-            : parameterOrVariable.label
+          const {description, label, name} = parameterOrVariable
+          const displayedDescription = description || label
           return (
-            <Link key={`${name}-${type}`} to={`/${name}`}>
+            <Link key={`${name}`} to={`/${name}`}>
```
MattiSG: > Why interpolate? Wouldn't `key={name}` work? 🤔

src/search.js
```diff
@@ -1,22 +1,28 @@
...
   if (isEmpty(normalizedQuery)) {
     return sortBy(prop('name'), parametersAndVariables)
   }
-  return sortWith([
-    ascend(item => weightIn(item.name, normalizedQuery)),
-    descend(item => item.normalizedDescription.includes(normalizedQuery)),
-    ascend(prop('name')),
-  ], parametersAndVariables)
+  const matchesInName = pipe(
+    filter(item => item.name.includes(normalizedQuery)),
```
MattiSG: Use [`splitWhen`](http://ramdajs.com/docs/#splitWhen) to avoid iterating twice on the input set :wink:

MattiSG:
There's an issue with the commit message for 6c74fa4.

src/components/pages/home.jsx
cbenz: Haha, this is dumb indeed. But after more thought I think this change was a mistake. It's *possible* as we said, that a variable and a parameter have the same name. That's why I added a `.type == 'variable' (or 'parameter')` attribute.

I must fix that.

src/search.js
cbenz: That was [`partition`](http://ramdajs.com/docs/#partition) actually (`splitWhen` stops partitioning at first match). But thanks for the good idea!

MattiSG:
Please don't force-push while in the review process. This makes it impossible to track changes since last review, thus making the reviewer's job very hard.

Issue #63, opened 2017-03-07
Add integration tests


MattiSG:
Cover the legislation explorer with tests that prove:

- The page loads and shows metadata on the loaded tax and benefit system.
- The search does filter the list.
- Navigating to an item of the list shows details.
- Going back keeps the search.

cbenz:
@MattiSG I chose [Watai](https://github.com/MattiSG/Watai) to implement integration tests.

I have a first error with my code in the `integration-tests` branch:

```sh
~/D/o/legislation-explorer (master|✚1…) $ watai tests
error: **Error in file "/home/cbenz/Dev/openfisca/legislation-explorer/tests/1 - AppScenario.js"** path=/home/cbenz/Dev/openfisca/legislation-explorer/tests/1 - AppScenario.js
debug: /home/cbenz/Dev/openfisca/legislation-explorer/tests/1 - AppScenario.js:5
    "AppComponent.countryPackageInfo": countryPackageNameAndDescription
                                       ^
ReferenceError: countryPackageNameAndDescription is not defined
    at /home/cbenz/Dev/openfisca/legislation-explorer/tests/1 - AppScenario.js:5:40
    at Object.exports.runInContext (vm.js:44:17)
    at loadScenario (/usr/local/lib/node_modules/watai/src/controller/SuiteLoader.js:348:7)
    at wrapper.extend.$owner (/usr/local/lib/node_modules/watai/node_modules/mootools/lib/mootools-core-1.4.5-server.js:997:23)
    at Function.<anonymous> (/usr/local/lib/node_modules/watai/node_modules/mootools/lib/mootools-core-1.4.5-server.js:320:44)
    at loadAllFiles (/usr/local/lib/node_modules/watai/src/controller/SuiteLoader.js:257:10)
    at wrapper.extend.$owner (/usr/local/lib/node_modules/watai/node_modules/mootools/lib/mootools-core-1.4.5-server.js:997:23)
From previous event:
    at getRunner (/usr/local/lib/node_modules/watai/src/controller/SuiteLoader.js:175:4)
    at wrapper.extend.$owner (/usr/local/lib/node_modules/watai/node_modules/mootools/lib/mootools-core-1.4.5-server.js:997:23)
    at Object.<anonymous> (/usr/local/lib/node_modules/watai/src/index.js:40:7)
/home/cbenz/Dev/openfisca/legislation-explorer/tests/1 - AppScenario.js:5
[...]

PR #65, opened 2017-03-13
Add integration tests

cbenz:
Connected to #63

### DIFF
--- a/tests/1 - AppScenario.js
+++ b/tests/1 - AppScenario.js
@@ -2,6 +2,7 @@ description: "The header should show metadata on the loaded tax benefit system."
 
 steps: [
   {
-    "AppComponent.countryPackageInfo": countryPackageNameAndDescription
+    "AppComponent.countryPackageInfo": countryPackageNameAndDescription,
+    "AppComponent.baseline": baseline
   }
 ]
--- a/tests/1 - AppScenario.js
+++ b/tests/1 - AppScenario.js
@@ -1,0 +1,7 @@
+description: "The header should show metadata on the loaded tax benefit system.",
+
+steps: [
+  {
+    "AppComponent.countryPackageInfo": countryPackageNameAndDescription
+  }
+]
--- a/tests/3 - NavigationScenario.js
+++ b/tests/3 - NavigationScenario.js
@@ -1,0 +1,13 @@
+description: "Navigating to an item of the list should show details, going back should keep the search.",
+
+steps: [
+  SearchBarComponent.searchFor(aahQuery),
+  AppComponent.goToFirstResult(),
+  {
+    "DetailsComponent.description": aahResultDescription
+  },
+  DetailsComponent.goBack(),
+  {
+    "SearchBarComponent.field": aahQuery
+  }
+]
--- a/tests/AppComponent.js
+++ b/tests/AppComponent.js
@@ -1,1 +1,2 @@
+baseline: "#baseline",
 countryPackageInfo: "#country-package-info"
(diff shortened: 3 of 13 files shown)


tests/1 - AppScenario.js
```diff
@@ -0,0 +1,8 @@
+description: "The header should show metadata on the loaded tax benefit system.",
```
fpagnoux: Not sure if it's a good practise to put "should" all the time. Affirmatives sentences are less heavy.

"The header displays metadata on the loaded tax and benefit system" ?
"Make sure the header displays metadata on the loaded tax and benefit system" ?

@MattiSG, what is your watai expert opinion ?

tests/3 - NavigationScenario.js
```diff
@@ -0,0 +1,13 @@
+description: "Navigating to an item of the list should show details, going back should keep the search.",
+
+steps: [
+  SearchBarComponent.searchFor(aahQuery),
+  AppComponent.firstResult(),
```
fpagnoux: Actions should be verbs.

tests/AppComponent.js
```diff
@@ -0,0 +1,4 @@
+baseline: "#baseline",
+countryPackageInfo: "#country-package-info",
+firstResultLink: '#search-results :first-child a',
```
fpagnoux: Why not simpler selector `#search-results a`, or even `ul a`?

tests/AppComponent.js
```diff
@@ -0,0 +1,4 @@
...
+firstResultName: '#search-results :first-child .name',
```
fpagnoux: Why bothering adding an extra `class="name"` instead of just using `#search-results h4`, or even `ul h4` ?

tests/AppFixture.js
```diff
@@ -0,0 +1,2 @@
+baseline = /\d+ variables et \d+ paramètres référencées pour modéliser le système socio-fiscal français\./
```
fpagnoux: It seems too rigid, we don't need to test so precisely.

`/\d+ variables et \d+ paramètres référencées/` would probably be enough and avoid reworking on the test if we change the second part a little.

fpagnoux: Also `référencées` -> `référencés`

tests/DetailsComponent.js
```diff
@@ -0,0 +1,2 @@
+description: "h1",
+goBackLink: "#go-back"
```
fpagnoux: > For links, you can use a [special selector](https://github.com/sgmap/mes-aides-ui/blob/master/test/integration/components/HomepageComponent.js).
> Here, it could be `goBackLink: { a: 'Voir' },`
> This would avoid introducing an id.

tests/SearchBarFixture.js
```diff
@@ -0,0 +1,3 @@
+aahQuery = "aah"
+aahResultName = "aah"
+aahResultDescription = "Allocation adulte handicapé (Individu) mensualisée"
```
fpagnoux: Too much details, this makes us sensitive to this (bad) description. `/Allocation adulte handicapé/` should be enough.

cbenz:
Thanks @fpagnoux ! Quoting [Watai](https://github.com/MattiSG/Watai/wiki/Tutorial#description):

> It is good practice to write a sentence with “should” as the main verb, so that failures and successes may both be meaningful.

tests/3 - NavigationScenario.js
cbenz: This relies on `firstResultLink` in AppComponent. The `firstResult` function is probably inferred. 

I'm going to rename the link to `goToFirstResultLink`.

tests/AppComponent.js
cbenz: Hmm this would select all the links nested in `#search-results`.

fpagnoux:
But please squash the reverts before merging.

Issue #64, opened 2017-03-07
Run integration tests in CI


MattiSG:
Run #63 in continuous integration.

PR #66, opened 2017-03-13
Run integration tests in CI

cbenz:
Connected to #64 

Follows #65

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,14 @@
+dependencies:
+  cache_directories:
+    - "~/cache"
+  pre:
+    # Only create directory if not already present (--parents option).
+    - mkdir --parents ~/cache
+  post:
+    # Only downloads if not already present (--no-clobber option)
+    - cd ~/cache && wget --no-clobber http://selenium-release.storage.googleapis.com/3.3/selenium-server-standalone-3.3.0.jar
+    - java -jar ~/cache/selenium-server-standalone-3.3.0.jar:
+        background: true
 deployment:
   production:
     branch: master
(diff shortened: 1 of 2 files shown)


cbenz:
It seems that chrome driver is installed by default in CircleCI containers.

circle.yml
```diff
@@ -1,3 +1,18 @@
...
+    - mkdir --parents ~/cache
+  post:
+    # Only downloads if not already present (--no-clobber option)
+    - cd ~/cache && wget --no-clobber http://selenium-release.storage.googleapis.com/3.3/selenium-server-standalone-3.3.0.jar
+    - java -jar ~/cache/selenium-server-standalone-3.3.0.jar:
+        background: true
+test:
+  pre:
+    - npm start:
+        background: true
```
MattiSG: I would recommend you wait for the server to start before starting the tests.

Suggested implementation:

```
   # wait for server
    - wget --retry-connrefused --waitretry=1 --output-document=/dev/null http://localhost:$PORT
```

circle.yml
```diff
@@ -1,3 +1,18 @@
+dependencies:
+  cache_directories:
+    - "~/cache"
+  pre:
+    # Only create directory if not already present (--parents option).
+    - mkdir --parents ~/cache
+  post:
+    # Only downloads if not already present (--no-clobber option)
+    - cd ~/cache && wget --no-clobber http://selenium-release.storage.googleapis.com/3.3/selenium-server-standalone-3.3.0.jar
+    - java -jar ~/cache/selenium-server-standalone-3.3.0.jar:
```
MattiSG: If you test only with Chrome on CI, you don't need to install the Selenium server. Just start `chromedriver`.

cbenz:
It seems that SauceLabs has only non-free pricing plans. I'm merging this PR to get it done but we may enhance that in the future.

MattiSG:
https://saucelabs.com/open-source

Issue #69, opened 2017-03-21
Load legislation parameters from the parameters v2 web API


MattiSG:
Get parameters list from the prerelease API endpoint.

The barèmes won't have values for the moment, that's ok.

cbenz:
The problem on which we spent a couple of hours with @fpagnoux is already described here: https://github.com/corydolphin/flask-cors/issues/147

I [commented](https://github.com/corydolphin/flask-cors/issues/147#issuecomment-289539799) to give a solution we found.

fpagnoux:
I reproduced the issue without `flask_cors`, and even without CORS header.  `flask_cors` is innocent.

PR #70, opened 2017-03-28
Use new param api

fpagnoux:
Connected to #69

### DIFF
--- a/config/development.js
+++ b/config/development.js
@@ -4,6 +4,7 @@
 
 const HOST = process.env.HOST || "localhost",
   apiBaseUrl = process.env.API_URL || `http://${HOST}:2000`,
+  parameterApiBaseUrl = process.env.FLASK_API_URL || `http://${HOST}:5000`,
   gitHubProject = "openfisca/openfisca-france",
   gitWebpageUrl = "https://github.com/openfisca/legislation-explorer",
   piwikConfig = null,
@@ -15,7 +16,8 @@ export default {
   apiBaseUrl,
   gitHubProject,
   gitWebpageUrl,
+  parameterApiBaseUrl,
   piwikConfig,
   useCommitReferenceFromApi,
-  websiteUrl
+  websiteUrl,
 }
--- a/src/components/pages/parameter-or-variable.jsx
+++ b/src/components/pages/parameter-or-variable.jsx
@@ -1,5 +1,5 @@
 import React, { PropTypes } from "react"
-import { isNil } from "ramda"
+import { isNil, isEmpty } from "ramda"
 import { Link, locationShape } from "react-router"
 
 import * as AppPropTypes from "../../app-prop-types"
@@ -7,6 +7,7 @@ import NotFoundPage from "./not-found"
 import Parameter from "../parameter"
 import Variable from "../variable"
 import {searchInputId} from "./home"
+import {fetchParameter} from "../../webservices"
 
 
 const ParameterOrVariablePage = React.createClass({
@@ -23,11 +24,20 @@ const ParameterOrVariablePage = React.createClass({
     parameters: PropTypes.objectOf(AppPropTypes.parameterOrScale).isRequired,
     variables: PropTypes.arrayOf(AppPropTypes.variable).isRequired,
   },
+  getInitialState() {
+    return {parameter: {}};
+  },
+  componentDidMount() {
+    fetchParameter(this.props.params.
+      name).then(
+      parameter => this.setState({parameter: parameter})
+    )
+  },
   render() {
     const { searchQuery, searchResults } = this.context
     const {countryPackageName, countryPackageVersion, currency, location, parameters, params, variables} = this.props
     const {name} = params
-    const parameter = parameters.find(parameter => parameter.name === name)
+    const {parameter} = this.state
     const variable = variables.find(variable => variable.name === name)
     if (isNil(parameter) && isNil(variable)) {
       return (
@@ -35,6 +45,13 @@ const ParameterOrVariablePage = React.createClass({
           location={location}
           message={`« ${name} » n'est ni un paramètre ni une variable d'OpenFisca.`}
         />
+      )
+    }
+    if (isEmpty(parameter) && isNil(variable)) {
+      return (
+        <div>
+          Loading...
+        </div>
       )
     }
     const goBackLocation = {
@@ -55,7 +72,7 @@ const ParameterOrVariablePage = React.createClass({
           }
         </Link>
         {
-          parameter && (
+          (! isEmpty(parameter)) && (
             <Parameter
               countryPackageName={countryPackageName}
               countryPackageVersion={countryPackageVersion}
--- a/src/components/pages/parameter-or-variable.jsx
+++ b/src/components/pages/parameter-or-variable.jsx
@@ -1,5 +1,5 @@
 import React, { PropTypes } from "react"
-import { isNil, isEmpty } from "ramda"
+import { isNil } from "ramda"
 import { Link, locationShape } from "react-router"
 
 import * as AppPropTypes from "../../app-prop-types"
@@ -26,17 +26,20 @@ const ParameterOrVariablePage = React.createClass({
   },
   getInitialState() {
     const variable = this.props.variables.find(variable => variable.name === this.props.params.name)
-    return {parameter: {}, variable: variable}
+    return {variable: variable, parameter: null, waitingForResponse: true}
   },
   componentDidMount() {
+    // If there is a variable matching the name, we don't need fetch the parameter API
     if (! this.state.variable) {
       fetchParameter(this.props.params.name).then(
         parameter => {
-          this.setState({parameter: parameter})
+          this.setState({parameter: parameter, waitingForResponse: false})
         }
       ).catch(
-        () => this.setState({parameter: undefined})
+        () => this.setState({waitingForResponse: false})
       )
+    } else {
+      this.setState({waitingForResponse: false})
     }
   },
   render() {
@@ -44,17 +47,19 @@ const ParameterOrVariablePage = React.createClass({
     const {countryPackageName, countryPackageVersion, currency, location, parameters, params, variables} = this.props
     const {name} = params
     const {parameter, variable} = this.state
+
+    if (this.state.waitingForResponse) {
+      return (
+        <p>Chargement des valeurs…</p>
+      )
+    }
+
     if (isNil(parameter) && isNil(variable)) {
       return (
         <NotFoundPage
           location={location}
           message={`« ${name} » n'est ni un paramètre ni une variable d'OpenFisca.`}
         />
-      )
-    }
-    if (isEmpty(parameter) && isNil(variable)) {
-      return (
-        <p>Chargement des valeurs…</p>
       )
     }
     const goBackLocation = {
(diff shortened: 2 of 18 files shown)


src/components/pages/parameter-or-variable.jsx
```diff
@@ -20,15 +21,29 @@ const ParameterOrVariablePage = React.createClass({
...
-    parameters: PropTypes.arrayOf(AppPropTypes.parameterOrScale).isRequired,
+    parameters: PropTypes.objectOf(AppPropTypes.parameterOrScale).isRequired,
     variables: PropTypes.arrayOf(AppPropTypes.variable).isRequired,
   },
+  getInitialState() {
+    const variable = this.props.variables.find(variable => variable.name === this.props.params.name)
+    return {parameter: {}, variable: variable}
+  },
+  componentDidMount() {
+    if (! this.state.variable) {
```
fpagnoux: Add comment

src/components/pages/parameter-or-variable.jsx
```diff
@@ -37,6 +52,11 @@ const ParameterOrVariablePage = React.createClass({
         />
       )
     }
+    if (isEmpty(parameter) && isNil(variable)) {
```
fpagnoux: Add comment

src/components/parameter.jsx
```diff
@@ -41,7 +41,13 @@ function getBracketsAtInstant(brackets, instant) {
 
 
 function getTodayInstant() {
-  return new Date().toJSON().slice(0, "YYYY-MM-DD".length)
+  return new Date().toISOString().slice(0, "YYYY-MM-DD".length)
+}
+
+
+function getDayBefore(day) {
+  // 86400000 ms is day
```
fpagnoux: is a day

src/components/parameter.jsx
```diff
@@ -317,7 +293,6 @@ const Parameter = React.createClass({
     const {format, unit} = parameter
     return (
       <div>
```
fpagnoux: Do we need this `div`?

fpagnoux: Do we need this `div` ?

src/search.js
```diff
@@ -27,7 +32,7 @@ export function addNormalizedDescription(propertyName, objects) {
   return map(
     object => assoc(
       'normalizedDescription',
-      has(propertyName, object)
+      has(propertyName, object) && object.propertyName
```
fpagnoux: should probably be object[propertyName]

fpagnoux:
I'm working on integration tests;

fpagnoux:
Ok, done with the tests and it's working (I must have had a bug with selenium server).

@cbenz there is a problem with the CIRCLE tests though: only one container pass, the 3 others fail with `The Selenium server could not be reached`. Will this be solve by #71 ?

cbenz:
@fpagnoux Yes, that's why I recommend merging #71 first.

src/components/pages/parameter-or-variable.jsx
```diff
@@ -20,15 +21,39 @@ const ParameterOrVariablePage = React.createClass({
     currency: PropTypes.string.isRequired,
     location: locationShape.isRequired,
     params: PropTypes.shape({ name: PropTypes.string.isRequired }).isRequired, // URL params
-    parameters: PropTypes.arrayOf(AppPropTypes.parameterOrScale).isRequired,
+    parameters: PropTypes.objectOf(AppPropTypes.parameterOrScale).isRequired,
     variables: PropTypes.arrayOf(AppPropTypes.variable).isRequired,
   },
+  getInitialState() {
+    const variable = this.props.variables.find(variable => variable.name === this.props.params.name)
+    return {variable: variable, waitingForResponse: true}
```
cbenz: Can you initialize `parameters` key to `null`?

cbenz: Rename `waitingForResponse` to `loading` (it's more commonly used)

src/components/parameter.jsx
```diff
@@ -41,7 +41,13 @@ function getBracketsAtInstant(brackets, instant) {
...
 
 function getTodayInstant() {
-  return new Date().toJSON().slice(0, "YYYY-MM-DD".length)
+  return new Date().toISOString().slice(0, "YYYY-MM-DD".length)
+}
+
+
+function getDayBefore(day) {
+  // 86400000 ms is a day
+  return new Date(Date.parse(day) - 86400000).toISOString().slice(0, "YYYY-MM-DD".length)
```
cbenz: What is `86400000`? Could you add a comment?

src/search.js
```diff
@@ -1,10 +1,15 @@
-import {ascend, assoc, concat, filter, has, isEmpty, map, pipe, prop, sortBy, sortWith, partition} from "ramda"
+import {ascend, assoc, concat, filter, isEmpty, map, pipe, prop, sortBy, sortWith, partition} from "ramda"
 import * as diacritics from 'diacritics'
 
+function preprocessParameters(parameters) {
+  return Object.keys(parameters).map(
+    paramId => Object.assign(parameters[paramId], {'name': paramId})
```
cbenz: Warning: this form of `Object.assign` mutates its first argument (`parameters[paramId]`).

Instead, use `Object.assign({}, parameters[paramId], {'name': paramId})`

Or use `assoc` from Ramda, like this: `assoc(parameters[paramId], {'name': paramId})` 😄

src/search.js
```diff
@@ -27,7 +32,7 @@ export function addNormalizedDescription(propertyName, objects) {
   return map(
     object => assoc(
       'normalizedDescription',
-      has(propertyName, object)
+      object[propertyName]
```
cbenz: > I'm not totally OK with this change: the test is about the presence of the key. Using `has` is more explicit about the intention than trying to access the key, which makes me wonder: will it raise an exception? Oh ... no, it's JavaScript, it will be... `undefined` ... or `null` ? On the other side, `has` returns a `boolean` which is reassuring!

src/components/pages/parameter-or-variable.jsx
MattiSG: > **META**
>
> In order to avoid frustration for all contributors, I would recommend to be more explicit about how important something is.
> 
> If it's only a matter of personal preference, then don't make your review sound as an imperative change in order to allow the merge. This can be done either through phrasing (“I'd feel more at ease with…”, “I'd prefer…”, “I'd like…”) or through quote markup.
> 
> If you really feel that this name is harmful, or is a large missed opportunity, then a reference would be welcome. I personally don't believe that `loading` is so much more common than `waitingForResponse`, for example :wink:

src/search.js
MattiSG: Actually, it's more complicated than that… `Boolean(object[propertyName])` ≠ `has(propertyName, object)`, because [`has`](http://ramdajs.com/docs/#has) only looks for an **own** property and thus doesn't walk up the prototype chain.

So this change changes two things: the return type and the property scope.
I am personally fine with it, since the value is immediately consumed, and is so unambiguously as a boolean value. There is thus no API change at all, so no risk.
As a JS person, I even prefer the native approach, as I know exactly what looking up a property entails, while I had to look up ramda's doc to discover that `has` was an alias to `hasOwnProperty` (i.e. no prototype chain walk).

This comment, as well as [this one](https://github.com/openfisca/legislation-explorer/pull/70#discussion_r108659248) on initialising properties are clearly a matter of style and of embracing falsies or feeling reassured by pure boolean values.

If we are to all work on this code at one point or another, we should agree on the tooling and the coding style, such as embracing ramda or trying to get away from it. Unless we decide to get away from that library, I don't think we should change code that works.

src/components/parameter.jsx
fpagnoux: There is a comment :D.
I'll make it a constant to make it clearer.

src/search.js
fpagnoux: This is not about code style.

If object **has** a `propertyName` property, but that the value of this property is `null`, then we dont wan't to remove the diacritics, as it would cause an error.

We need this change because with the new API, a parameter description may be null.

src/components/pages/parameter-or-variable.jsx
fpagnoux: What would be the point ? Is this a common practise to initialise states properties to null in React ?

src/search.js
fpagnoux: 👍

src/components/parameter.jsx
MattiSG: I usually solve that problem by writing `24 * 60 * 60 * 1000`.

src/search.js
MattiSG: My bad, I didn't make the link with the new API. Thanks a lot for the clarification @fpagnoux!  :)

src/components/pages/parameter-or-variable.jsx
cbenz: It's not related to React: I'd say it's about following a principle of least surprise, minimizing the number of places I have to look at when I want to know what properties an may object have. But don't change anything, indeed in JS it's common to add properties after the object is created.

MattiSG: Using the initial state as the “shape” or “schema” of the full state indeed sounds like a proper reason for adding a key.

src/components/parameter.jsx
```diff
@@ -114,14 +120,12 @@ const Parameter = React.createClass({
...
           </div>
-          {this.renderDefinitionsList(parameter)}
-          <hr/>
           <div className="row">
             <div className="col-lg-8">
               {
-                type === "Parameter"
-                  ? this.renderParameter(parameter)
-                  : this.renderScale(parameter)
+                type === "Scale"
```
cbenz: I don't think that this test is relevant anymore with the new API, which removes the `parameter['@type']` property (cf line above `const type = parameter["@type"]`)

config/development.js
```diff
@@ -4,6 +4,7 @@
 
 const HOST = process.env.HOST || "localhost",
   apiBaseUrl = process.env.API_URL || `http://${HOST}:2000`,
+  parameterApiBaseUrl = process.env.NEW_API_URL || `http://${HOST}:5000`,
```
fpagnoux: Done

fpagnoux: 👍

src/webservices.js
```diff
@@ -8,21 +8,27 @@ function fetchJson(url, options) {
...
 
 
-async function parseJsonResponse(response) {
-  const data = await response.json()
-  if (response.status >= 200 && response.status < 300) {
-    return data
-  } else {
-    throw new Error(JSON.stringify(data.error))
-  }
+function parseJsonResponse(response) {
```
fpagnoux: ¯\_(ツ)_/¯

fpagnoux:
@cbenz we didn't test with the same API !

I improved the robustness so that we don't have any error when https://github.com/openfisca/openfisca-web-api/pull/116 is merged, but I chose to not render scale parameters in this PR.

I think having a temporary solution to display the simple parameters via the flask API and the scales via the old one is too much work, as I assume we probably will adapt the scales rendering to the flask API in the next iteration.

src/webservices.js
cbenz: That was taken from one of the many tutorials about JS Promises... I reckon it's over-hype, but that's for better composability I think....

cbenz:
@fpagnoux You may now rebase on `master` to see integration tests happen in SauceLabs!

fpagnoux:
Considering that:
- I could not reproduce the bug reported
- @cbenz didn't encounter it
- The travis integration tests are all ok
- The reported bug indicates that the whole app is not working

I'm dismissing the review and moving forward.

Issue #68, opened 2017-03-21
Continuously test the cross-browser compatibility


MattiSG:
Use [SauceLabs](https://github.com/MattiSG/Watai/wiki/Set-up-parallel-continuous-testing).

MattiSG:
Fixed via #71.

PR #71, opened 2017-03-28
Continuously test the cross-browser compatibility

cbenz:
Connected to #68


cbenz:
@MattiSG After a couple of hours of investigation, I kept only the `browserName` driver capability, and replaced `iPhone` by `android`, which works fine :)

circle.yml
```diff
@@ -1,17 +1,48 @@
+# This config uses parallelism. See https://circleci.com/docs/1.0/environment-variables/#parallelism
 machine:
   environment:
-    # Use the same port as Selenium server.
-    DRIVER_PORT: 4444
-    # Use the same base URL as Selenium server.
-    DRIVER_BASE_URL: wd/hub
+    browser_name: $(case $CIRCLE_NODE_INDEX in 0) browser_name='chrome' ;; 1) browser_name='firefox' ;; 2) browser_name='internet explorer' ;; 3) browser_name='android' ;; esac; echo $browser_name)
```
fpagnoux: Could we move this complex logic to the `config-ci.js` file ?

MattiSG: Wow, nice, hadn't thought about this!! But it could indeed work with `process.env`  😃

cbenz: I'm OK to try right now ;)

MattiSG: Doing that is experimental: we're not sure there would be actual added value by hiding the complexity of a setup specifically done in CI in some other place than the CI config itself. Please timebox the experiment  :)

circle.yml
```diff
@@ -1,17 +1,48 @@
+# This config uses parallelism. See https://circleci.com/docs/1.0/environment-variables/#parallelism
```
MattiSG: > I don't find this comment to be very useful. You get `parallel: true` everywhere, I guess you would look up `parallel` in CircleCI if you didn't know what it was. At least, rephrase “uses parallelism” to “builds across multiple containers in parallel” or some other description that doesn't repeat the name without defining it :wink:

circle.yml
```diff
@@ -1,17 +1,48 @@
...
 machine:
   environment:
-    # Use the same port as Selenium server.
-    DRIVER_PORT: 4444
-    # Use the same base URL as Selenium server.
-    DRIVER_BASE_URL: wd/hub
+    browser_name: $(case $CIRCLE_NODE_INDEX in 0) browser_name='chrome' ;; 1) browser_name='firefox' ;; 2) browser_name='internet explorer' ;; 3) browser_name='android' ;; esac; echo $browser_name)
+    SAUCE_LOG: $CIRCLE_ARTIFACTS/sauce_log.txt
+    WATAI_TESTS: tests
+    SAUCE_USERNAME: cbenz
```
MattiSG: Please create an `openfisca` user. OpenSauce plans are offered per project, not per human user.

circle.yml
```diff
@@ -1,17 +1,48 @@
...
+    # SAUCE_ACCESS_KEY is defined in the CircleCI web UI
+    PORT: 2030 # Legislation Explorer ExpressJS server port
+
+dependencies:
+  post:
+    - npm install saucelabs # allow sending Watai results to SauceLabs
```
MattiSG: Are test results actually updated on SauceLabs with this version? Last time we checked some did but some didn't.

circle.yml
```diff
@@ -1,17 +1,48 @@
...
+    - tar -xzf sc-latest-linux.tar.gz
+    - npm run build
+
 test:
   pre:
-    - chromedriver --port=$DRIVER_PORT --url-base=$DRIVER_BASE_URL:
-        background: true
-    # Wait for server.
-    - wget --retry-connrefused --waitretry=1 --output-document=/dev/null http://localhost:$DRIVER_PORT/$DRIVER_BASE_URL/status
+    # start server
```
MattiSG: > I see no added value in this comment above `npm start`.

tests/config-ci.js
```diff
@@ -1,11 +1,25 @@
...
+
+var browserName;
+switch (process.env.CIRCLE_NODE_INDEX) {
+  case 0:
+    browserName = 'chrome';
+  case 1:
+    browserName = 'firefox';
+  case 2:
+    browserName = 'internet explorer';
+  case 3:
```
MattiSG: > I puked. Please, use an array or an object 😅

tests/config-ci.js
```diff
@@ -1,11 +1,25 @@
...
+
 module.exports = {
   seleniumServerURL: {
     hostname: 'ondemand.saucelabs.com',
     port: 80
   },
   driverCapabilities: {
-    'tunnel-identifier': 'circle-' + process.env.CIRCLE_BUILD_NUM + '-' + process.env.CIRCLE_NODE_INDEX
+    'tunnel-identifier': 'circle-' + process.env.CIRCLE_BUILD_NUM + '-' + process.env.CIRCLE_NODE_INDEX,
+    'browserName': browserName
```
MattiSG: > No need to quote the property name. `tunnel-identifier` has to only because it contains a dash.

tests/config-ci.js
cbenz: I think it's time to go to ~sleep~ beer (?) :-)

> #### META
>
> To avoid frustration of the team please do not puke!

circle.yml
MattiSG: Looking at the build log, I see no SauceLabs job URL at the end. This is a sign that the `saucelabs` module isn't working  :(

circle.yml
```diff
@@ -1,17 +1,47 @@
+# This config uses parallelism. See https://circleci.com/docs/1.0/environment-variables/#parallelism
 machine:
   environment:
-    # Use the same port as Selenium server.
-    DRIVER_PORT: 4444
-    # Use the same base URL as Selenium server.
-    DRIVER_BASE_URL: wd/hub
+    SAUCE_LOG: $CIRCLE_ARTIFACTS/sauce_log.txt
+    WATAI_TESTS: tests
+    SAUCE_USERNAME: cbenz
```
MattiSG: _(repeating due to previous comment incorrectly marked as outdated)_

Please create an `openfisca` user. OpenSauce plans are offered per project, not per human user.

tests/config-ci.js
```diff
@@ -0,0 +1,20 @@
+// this file is for use in CircleCI continuous integration environment
+
+var browserNames = ['chrome', 'firefox', 'internet explorer', 'android'];
+var browserName = browserNames[process.env.CIRCLE_NODE_INDEX];
```
MattiSG: > I don't think this intermediate variable is useful.

tests/config-ci.js
```diff
@@ -0,0 +1,20 @@
+// this file is for use in CircleCI continuous integration environment
+
+var browserNames = ['chrome', 'firefox', 'internet explorer', 'android'];
```
MattiSG: > This is so great ❤️
> I'm updating Watai's wiki.

circle.yml
MattiSG: Aaaah I get it: the `config-ci` is actually moved to `.watai/config`, which has lower precedence than `config`. The `views` are thus ignored… but I'm afraid that the `browser` instructions could be too!

tests/config-ci.js
MattiSG: Actually, precedence rules have to be double-checked (see https://github.com/openfisca/legislation-explorer/pull/71#discussion_r108738852).

cbenz:
I think this works now. But Watai results are not sent to SauceLabs. I tried with `npm install saucelabs@0.1.1` but it didn't resolve the problem. I don't know how to investigate; it's not blocking since Circle returns correctly the results of the tests to the PRs.

MattiSG:
> I don't know how to investigate

I explained why this happens in https://github.com/openfisca/legislation-explorer/pull/71#discussion_r108738852. The `config` file takes precedence as it should, and the views are thus not merged. Easy solution is to remove the specification of views in `config.js` in order to let `config-ci` define it.

MattiSG:
I don't get it: you said you were forcing but didn't. Reviewing again.

tests/config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
     baseURL: 'http://localhost:2030',
-    browser: 'chrome'
+    browser: 'firefox',
+    views: [ 'Verbose' ]
```
MattiSG: Remove this line to send test results to SauceLabs in CI.

tests/config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
     baseURL: 'http://localhost:2030',
```
MattiSG: > Add a `name: "OpenFisca legislation explorer"` property to increase readability in SauceLabs logs.

MattiSG:
I'm handling changes, will be faster.

MattiSG:
Before my review:

capture d ecran 2017-03-31 a 17 52 38

With my latest changes:

capture d ecran 2017-03-31 a 17 52 32

cbenz:
Thanks @MattiSG !

Indeed I was about to force, then I read again your message in which you asked for a ping first, so I preferred doing so :)

Issue #73, opened 2017-04-05
Display scale parameters in legislation explorer


fpagnoux:
Right now they have been [deactivated](https://legislation.openfisca.fr/impot_revenu.bareme)

UI needs to be discussed.

Propositions : 
- Select field to choose the date (with all the dates where the law has changed)
- Slider to choose the date
- Date picker
- All values displayed in collapsable blocks
- A graphical plot

fpagnoux:
### Decision:

We display all the values in the page, in an accordion if necessary.

PR #75, opened 2017-04-10
Render scales

### LINKED ISSUE #74: A value of 0 is presented as a removal from legislation
On https://legislation.openfisca.fr/impot_revenu.autre.finpfl

image

The API returns a boolean: https://api-test.openfisca.fr/parameter/impot_revenu.autre.finpfl

But [this condition](https://github.com/openfisca/legislation-explorer/blob/590bb9969db4ab2df8f30fe3a2bf5b234d48f2ab/src/components/parameter.jsx#L430) uses `! value` abusively, which matches the `false`.

> Meta: @fpagnoux @MattiSG What could we do to avoid this from now?

cbenz:
Connected to #73 
Fixes #74 

The fix of #74 is needed for this PR, it is not a "two-features" PR ;-)


cbenz:
@fpagnoux @MattiSG I'm not sure about the best way to test this PR with Watai. I'm open to propositions.

fpagnoux:
Displays:

image

fpagnoux:
I find it difficult to distinguish the different groups.

Suggestion: make the bottom border more thick and dark like this: 

image

fpagnoux:
> I'm not sure about the best way to test this PR with Watai

What is the difficulty you are encountering ? Testing the table ?

cbenz:
@fpagnoux I'd like to have some suggestions about which new Watai test to write.

fpagnoux:
🚨 **FORCE PUSH DETECTED DURING REWIEW**🚨

cbenz:
New rendering after my new commit:

image

src/components/scale.jsx
```diff
@@ -0,0 +1,57 @@
...
+                  <tr
+                    className={classNames({"first-bracket": thresholdIndex === 0})}
+                    key={`${startDateIndex}-${thresholdIndex}`}
+                  >
+                    {
+                      thresholdIndex === 0 && (
+                        <td rowSpan={Object.keys(bracketAtDate).length}>
+                          À partir du
+                          {" "}
+                          <FormattedDate value={startDate} />
```
fpagnoux: This seems equivalent to `À partir du <FormattedDate value={startDate} />`, that I find clearer. Is it a good practise to use `{" "}` ?

src/components/scale.jsx
```diff
@@ -0,0 +1,57 @@
...
+                        </td>
+                      )
+                    }
+                    <td>
+                      Au-dessus de
+                      {" "}
```
fpagnoux: Same than above.

src/assets/style.css
```diff
@@ -1,3 +1,7 @@
+tr.first-bracket:not(:first-child) > td {
+  border-top-width: 5px !important; /* Overrides ".table-bordered > tbody > tr > td" rule. */
```
fpagnoux: On Chrome 56, the `!important` does not seem necessary when we have `:not(:first-child)`.

Are we sure this is necessary ?

benjello:
I don't know if it is the right place to make the following remarks but I put the here anyway:
 - brackets should be displayed as follows: Entre 0 et 9700 €
 - Rate should represented as percentage values i.e  5,5 %

fpagnoux:
Some afterthoughts: 

- Like @benjello, `Entre 0.0 et 9700.0` seems clearer to me than `Au dessus de 0.0`. While `À partir du` is clear and simple for date ranges, I'm not sure it's limpid for marginal tax brackets (which are often misunderstood by many people).
- I'd also consider adding a header line to explicit what these values _are_ .

| Date  | Tranche | Taux marginal |
| ------------- | ------------- | ---|

fpagnoux:
About testing, following the spirit of the current tests, I'd have the following scenario : 
- Go to the `/impot_revenu.bareme` page
- Check the headers (if we choose to add them)
- For the first two date blocks
  - Check that the date cell has a row span, and is properly formatted
  - Check that the first two thresholds are properly formatted
  - Check that the first two values are properly formatted.

If there is in `france` a case of a marginal scale that doesn't exist anymore (@benjello ?) , I'd also test its first line.

I would *not* test the values of the row span, rates and threshold, as this would make our tests break at any parameter update. This is a hole in our test net, but the cost is too high (except if one day we decide to run the tests on a dummy country).

fpagnoux:
Finally, @benjello, we decided to postpone unit management in the the new api / legislation explorer.

src/app-prop-types.js
```diff
@@ -3,50 +3,14 @@ import {PropTypes} from "react"
...
-  ),
   description: PropTypes.string,
-  unit,
-  // Introspection (optional: values can be generated programmatically)
-  xml_file_path: PropTypes.string,
-  start_line_number: PropTypes.number,
-  end_line_number: PropTypes.number,
+  normalizedDescription: PropTypes.string,
+  brackets: PropTypes.objectOf(values),
+  values,
```
cbenz: yes http://es6-features.org/#PropertyShorthand

src/components/parameter.jsx
```diff
@@ -55,382 +19,42 @@ const Parameter = React.createClass({
...
-      instant,
-      instantText: intl.formatDate(instant),
-    })
+    parameter: AppPropTypes.parameter.isRequired,
   },
   render() {
     const {parameter} = this.props
-    const {brackets, description} = parameter
-    const descriptionMessage = description || "Aucune description"
+    const description = parameter.description|| "Aucune description"
```
cbenz: You're right! I'm adding http://eslint.org/docs/rules/space-infix-ops

src/components/value.jsx
```diff
@@ -0,0 +1,17 @@
...
+    children: AppPropTypes.value.isRequired,
+  },
+  render() {
+    const {children} = this.props
+    return <samp>{String(children)}</samp>
+  }
+})
+
+
+export default Value
```
cbenz: In this case, the `<Value>` component abstracts the transformation to string of a value (see [this line](https://github.com/openfisca/legislation-explorer/pull/75/files/fc53ba691d28d409e1d2b2288501baa46626383d#diff-8454535cc87db40d599490dc56ac44b4R12)). I would not like to repeat the `<samp>{String(value)}</samp>` code.

I think here we can use the [stateless functional components syntax](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components) (or even a normal factory function but that's not classic).

From a conceptual point of view a component is just a function and "propTypes" are type validation about function parameters.

By the way [propTypes are deprecated](https://facebook.github.io/react/blog/#new-deprecation-warnings) since React 15.5 but I don't want to upgrade now.

src/components/scale.jsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+  render() {
+    const {brackets} = this.props.parameter
+    return (
+      <table className="table table-bordered table-hover table-striped">
+        <tbody>
+          {
+            Object.keys(brackets).sort().reverse().map((startDate, startDateIndex) => {
+              const bracketAtDate = brackets[startDate]
+              return Object.keys(bracketAtDate).map((threshold, thresholdIndex) => {
```
cbenz: Is sort guaranteed by the new API?

src/components/scale.jsx
cbenz: Yes because otherwise the space will not be in the DOM. That's a limitation of the React virtual DOM engine (see [this issue](https://github.com/facebook/react/issues/1643)).

The other way to avoid the problem is to use a `<span>` but this produces a heavier DOM.

cbenz:
> So far, we are not handling units in the new version of the API and therefore in the legislation explorer.

@fpagnoux But if you suggest adding a header with "Taux marginaux", why not adding a `%` unit?

cbenz:
@fpagnoux @benjello Is this rendering OK for you (I took it from a random website)?

image

For the first bracket, if we say `Entre 0.0 et 9700.0`, I think `9700` is included.

But for the following brackets, would it be `Entre 9701 et 26791`?

Is it worth doing the `+ 1` operation?

---

Given this data from [the API](https://api-test.openfisca.fr/parameter/impot_revenu.bareme):

```json
{
    "0.0": 0.0,
    "9700.0": 0.14,
    "26791.0": 0.3,
    "71826.0": 0.41,
    "152108.0": 0.45
}
```

- the first key (`0.0`) is a starting value of the first bracket
- the following keys (`9700.0`, [...], `152108.0`) are stop values of the previous bracket

I didn't see it explained explicitly in https://github.com/openfisca/openfisca-web-api/issues/107 so I wanted to make it clear.

cbenz:
My last commit renders like that:

image

Is is OK?

src/app-prop-types.js
fpagnoux: Thanks 👍

fpagnoux:
I would not add the `+1`, I find it misleading.
 
Conceptually you pay `0%` for what is _under_ `9700` and `14%` for what is _above_ `9700`.

`9701` only plays a role because the income to tax is rounded before applying the scale. If we do not round, which may be the case for some scales, it doesn't make sense: with 9700.5, I'm already in the next bracket.

Also, when I read `Entre 9701 et...`, I understand that my 9701th euro won't be taxed, which is not the case.

src/components/value.jsx
fpagnoux: Thanks for the references.

What I still don't understand is why we have to use `String` in the first place. I tried locally to replace <Value> by <samp>, and it seems to render strictly the same way. Am I missing something ?

src/components/scale.jsx
fpagnoux: Something like 
```js
Object.keys(bracketAtDate).sort((a,b) => Number(a) -  Number(b))
```
 should avoid any double conversion.

The API indeed guarantees that the JSON is in the right order, but (@MattiSG tell me if I'm wrong) we don't have any guarantee that the JS **client** will preserve this order. For instance, `Object.keys(bracketAtDate)` could possibly take the keys in an arbitrary order.

I really don't know if this is just a theoretical risk (as most browsers usually preserve the key order), but I'm pretty sure it's not guaranteed in the spec. We are partially covered by the integration tests, but I'm not sure I want to take this risk.

src/assets/style.css
```diff
@@ -1,3 +1,7 @@
+tr.first-bracket:not(:first-child) > td {
+  border-top-width: 5px; /* Overrides ".table-bordered > tbody > tr > td" rule. */
```
fpagnoux: Now that we don't have the `!important`, do we still need the comment ?

fpagnoux:
>> So far, we are not handling units in the new version of the API and therefore in the legislation >explorer.
>
>@fpagnoux But if you suggest adding a header with "Taux marginaux", why not adding a `%` unit?

You're right. As we are in the case of a marginal scale, we _know_ the unit is percent, so 👍  for me.

src/components/scale.jsx
fpagnoux: >I think the space issue only comes when we split the code over several lines.
So `À partir du <FormattedDate value={startDate} />` should work fine, no ?

>I understand that `{" "}` is necessary when have long lines, but in this case, between two tiny lines, I think the cost overcome the benefits.

PR #81, opened 2017-05-03
Use the flask API to display variables

### LINKED ISSUE #77: Show definition period for variables
As an _analyst_,
I can read the definition period of a variable,
So that I know on which periods I can call it

- - - -

- Consume the data provided by openfisca/openfisca-web-api#125.
- Remove all references to the former API.
- Make the “definition period” header a link to the periods documentation page.

fpagnoux:
Fixes #77


src/components/pages/home.jsx
```diff
@@ -126,13 +126,12 @@ const SearchResults = React.createClass({
     return (
       <List items={items} type="unstyled">
         {parameterOrVariable => {
-          const {description, itemType, label, name} = parameterOrVariable
-          const displayedDescription = itemType === 'parameter' ? description : label
+          const {description, itemType, name} = parameterOrVariable
```
MattiSG: > `item` would probably be a better name.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-    const {formula, label, source_code} = variable
-    const labelMessage = label || `${variable.name} (aucun label)`
     return (
-      <DocumentTitle title={`${labelMessage} - Explorateur de la législation`}>
+      <DocumentTitle title={`${variable.id} - Explorateur de la législation`}>
         <div>
-          <div className="page-header">
-            <h1>{labelMessage}</h1>
+          <header className="page-header">
+            <h1>{variable.id}</h1>
```
MattiSG: > `<code>`  (within `<h1>`)? Or style `.page-header h1 { font-family: monospace; }`?

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
+            <p>{variable.description}</p>
+          </header>
+          {this.renderVariableMetadata(variable)}
+          <div>
+            <ExternalLink href={variable.source}>
+              Voir le code source de cette variable
```
MattiSG: > “Voir le” not very useful.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
+            </ExternalLink>
+          </div>
+          {variable.formulas && this.renderFormulas(variable.formulas)}
+          <div>
+            <ExternalLink href={`${config.apiBaseUrl}/variable/${variable.id}`}>
+              Voir la donnée brute au format JSON
```
MattiSG: > “Voir la” not very useful.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-        <span>
-          Formule de calcul jusqu'au
-          {" "}
-          <FormattedDate value={formula.stop_instant} />
-        </span>
-      )
+    const definitionPeriodMessage = {
+      YEAR: "de l'année",
+      MONTH: "du mois",
+      ETERNITY: "de l'éternité (cette variable ne peut pas évoluer avec le temps",
```
MattiSG: Missing closing paren.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-                          }
-                        }
-                      }
-                    </List>
-                  </dd>
-                )
-              }
-            </dl>
+          variable.references && (
+            <span>
```
MattiSG: An inline `<span>` cannot contain a block-level element `<dt>` / `<dd>`. Within a `<dl>`, only `<dt>` / `<dd>` children are allowed. This `<span>` doesn't seem useful at all, at least for the HTML…?

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
+              <dt>Références :</dt>
+              <dd><ul>
+                {
+                  variable.references.map((reference, idx) =>
+                    <li key={idx}>
```
MattiSG: What is this `key` attribute?

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-    const {parameters} = this.props
-    const inputVariableNames = formula.input_variables
-    const formulaParameterNames = formula.parameters
     return (
       <div>
+        <dl>
+          <dt>Cette variable est définie pour</dt>
+          <dd>{entityMessage[variable.entity]}.</dd>
+          <dt>Elle se calcule à l'échelle</dt>
+          <dd>{definitionPeriodMessage[variable.definitionPeriod]}</dd>
```
MattiSG: Inconsistent punctuation: add full stops for each `<dd>`.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
+      ETERNITY: "de l'éternité (cette variable ne peut pas évoluer avec le temps",
     }
-  },
-  renderFormula(formula) {
-    const {parameters} = this.props
-    const inputVariableNames = formula.input_variables
-    const formulaParameterNames = formula.parameters
     return (
       <div>
+        <dl>
```
MattiSG: > Suggested styling (needs a classname): `dt, dd { float: left; } dt { clear: both; }`.

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-                        }
-                      }
-                    </List>
-                  </dd>
-                )
-              }
-            </dl>
+          variable.references && (
+            <span>
+              <dt>Références :</dt>
```
MattiSG: `&nbsp;`

src/components/variable.jsx
```diff
@@ -43,275 +15,110 @@ const Variable = React.createClass({
...
-                  {entry => `${entry[0]} = ${entry[1]}`}
-                </List>
-              </dd>
+                {stopDate && (! startDate) &&
+                  <h3>Jusqu'au <FormattedDate value={stopDate} />&nbsp;:</h3>
+                }
+                {startDate && stopDate &&
+                  <h3>Du <FormattedDate value={startDate} /> au <FormattedDate value={stopDate} />&nbsp;:</h3>
+                }
+                <pre>{formulas[date].content}</pre>
```
MattiSG: Proper markup would be `<pre><code>`.

src/periods.js
```diff
@@ -0,0 +1,6 @@
+function getDayBefore(day) {
+  const ONE_DAY = 86400000 // in ms
+  return new Date(Date.parse(day) - ONE_DAY).toISOString().slice(0, "YYYY-MM-DD".length)
+}
+
+export default getDayBefore
```
MattiSG: > These modifiers could be added to the function declaration itself.

src/webservices.js
```diff
@@ -4,31 +4,37 @@ import config from "./config"
 
 
 function fetchJson(url, options) {
-  return fetch(url, options).then(parseJsonResponse)
+  return fetch(url, options)
+    .then(response => response.json()
+      .then(data => {
```
MattiSG: The whole point of promises is that they can be chained. `return` the JSON and use `then` on the return value of `then`, not on `response.json()`.

[See more](http://stackoverflow.com/questions/22944717/callback-pyramid-even-with-when-promise).

src/webservices.js
```diff
@@ -4,31 +4,37 @@ import config from "./config"
...
+        if (response.status >= 200 && response.status < 300) {
+          return {
+            data,
+            'country-package': response.headers.get('country-package'),
+            'country-package-version': response.headers.get('country-package-version'),
+          }
+        } else {
+          throw new Error(JSON.stringify(data.error))
```
MattiSG: `data.error` might be empty if `response.status` is a 3XX. Need to handle that case.

src/components/variable.jsx
fpagnoux: React imposes to have an "enclosing tag". We can get rid of it by un-factorizing the code I guess.

src/components/variable.jsx
cbenz: This is specific to React: https://facebook.github.io/react/docs/lists-and-keys.html#keys

The key attribute helps re-rendering the DOM in an optimal way:

> Keys help React identify which items have changed, are added, or are removed.

src/webservices.js
fpagnoux: I know but I need the `response` object !

src/components/variable.jsx
fpagnoux: Didn't manage to make it look good. To be done when we redesign the page.

PR #88, opened 2017-05-31
Document the preview API in the legislation explorer

fpagnoux:
Connected openfisca/openfisca-core#514

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -44,6 +44,9 @@ const App = React.createClass({
       searchResults: findParametersAndVariables(parameters, variables, searchQuery),
     }
   },
+  isCurrentRoute(route) {
+    return this.props.location.pathname == route
+  },
   render() {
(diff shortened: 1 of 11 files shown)


src/components/app.jsx
```diff
@@ -44,6 +44,9 @@ const App = React.createClass({
       searchResults: findParametersAndVariables(parameters, variables, searchQuery),
     }
   },
+  isCurrentRoute(route) {
+    return this.props.location.pathname == route
```
cbenz: This works in this case because routes have no param, but I think the idiomatic way to match a path is with [`matchPath`](https://reacttraining.com/react-router/web/api/matchPath) :

```js
matchPath(route, this.props.location)
```

Since I'm not sure could you just try?

MattiSG:
Well, that review was quick  ;)  #89.

MattiSG:
And the whole page reloads if I switch the tab 😕

PR #94, opened 2017-06-15
Upgrade webpack

fpagnoux:
Problem encountered trying to upgrade to `wepack2`: 

I start the server with `npm run dev:prod-api`. 

I get a few warnings, but the server starts correctly:
```
DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56.
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
[0] Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
[0] Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
```
However, when I open a page, I get the error in the console:

```
yaml.js:263 Uncaught Error: Cannot find module "."
    at webpackMissingModule (yaml.js:263)
    at Object.defineProperty.value (yaml.js:263)
    at __webpack_require__ (bootstrap b3622f4…:659)
    at fn (bootstrap b3622f4…:85)
    at r (swagger-ui.js:1)
    at Object.<anonymous> (swagger-ui.js:1)
    at __webpack_require__ (bootstrap b3622f4…:659)
    at fn (bootstrap b3622f4…:85)
    at Object.<anonymous> (swagger.jsx:2)
```
And the JS of the page doesn't work properly.

`yaml.js` is a sub-sub dependency, that we probably don't use at all 😞 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "type": "git",
     "url": "https://github.com/openfisca/legislation-explorer.git"
   },
-  "author": "Christophe Benz <christophe.benz@data.gouv.fr>",
+  "author": "OpenFisca <contact@openfisca.fr>",
   "license": "AGPLv3+",
   "dependencies": {
     "babel-cli": "^6.11.4",
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "description": "OpenFisca - Legislation Explorer",
   "scripts": {
-    "build": "webpack --config ./webpack.config.prod.babel.js",
-    "build:tunisia": "COUNTRY_PRODUCTION_CONFIG=tunisia webpack --config ./webpack.config.prod.babel.js",
+    "build": "webpack --config ./webpack.config.prod.babel.js --progress",
+    "build:tunisia": "COUNTRY_PRODUCTION_CONFIG=tunisia webpack --config ./webpack.config.prod.babel.js --progress",
     "clean": "rm -rf public",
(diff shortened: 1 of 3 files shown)


fpagnoux:
[This issue](https://github.com/swagger-api/swagger-ui/issues/3046) seems to describe the same problem, but the suggested fix, using `swagger-ui-dist`, doesn't work as we get an error `Error: only one instance of babel-polyfill is allowed`.

shockey:
Hey @fpagnoux! We merged https://github.com/swagger-api/swagger-ui/pull/2959 yesterday, so our upcoming Swagger-UI release should resolve your babel-polyfill problem.

MattiSG:
@fpagnoux who is this benefiting to? 🤔

MattiSG:
Nice, thanks @shockey! 😃

fpagnoux:
Thanks 🙂  !

fpagnoux:
> @fpagnoux who is this benefiting to? 🤔

This would have resolved issues encountered earlier, but as upgrading to webpack 2 didn't work, we found an alternative. 
I'm closing this PR, to be continued if one day we do need to upgrade.

Issue #72, opened 2017-03-29
Add an international description


MattiSG:
Include the following text, above the count of parameters and variables, with language depending on the browser language:

## `en`

> OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.

## `fr`

> OpenFisca rend le droit calculable. Il s’agit d’un moteur de calcul libre et ouvert qui permet de modéliser le système socio-fiscal de manière collaborative et transparente. Ce logiciel est utilisé par des chercheurs en économie et des services publics.

MattiSG:
The browser support should be:

- IE11
- Edge
- Safari 9
- Firefox 43
- Chrome 52

PR #96, opened 2017-06-21
Internationalise site header

sandcha:
Connected to #72 

Translate site header message according to browser language.
(Set default language to french and define translations in json files.)


fpagnoux:
✅  I get the english version, if I set my browser language to English.

src/components/app.jsx
```diff
@@ -65,8 +80,12 @@ const App = React.createClass({
                 </div>
                 <div className="col-md-9">
                   <p id="baseline">
-                    {Object.keys(variables).length} variables et {Object.keys(parameters).length} paramètres
-                    référencés pour modéliser le système socio-fiscal français.
+                    <FormattedMessage {...messages.header} />
+                    <br/>
```
fpagnoux: Multiplying `<br/>` to increase the distance between 2 texts blocks is not great.
Maybe each should be a `p` and we could set a margin in css ?

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+import en from "react-intl/locale-data/en"
+
+const DEFAULT_LANGUAGE = 'fr'
+
+export function loadTranslations(langDir){
+  addLocaleData(...fr)
+  addLocaleData(...en)
+
+  var messages = new Map()
+  var dotIndex = -1
```
fpagnoux: Why initiate this value here ? You only need it locally in the function line 22.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+import fr from "react-intl/locale-data/fr"
+import en from "react-intl/locale-data/en"
+
+const DEFAULT_LANGUAGE = 'fr'
+
+export function loadTranslations(langDir){
+  addLocaleData(...fr)
+  addLocaleData(...en)
+
+  var messages = new Map()
```
fpagnoux: Why using a `Map` rather than a more commonly used object (`{}`) ?

src/assets/lang/en.json
```diff
@@ -0,0 +1,4 @@
+{
+  "header" : "OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
+  "stats" : "{nbv, number} variables and {nbp, number} parameters referenced to modelize the french tax-benefit system."
+}
```
MattiSG: Beyond GitHub, this is actually better for diff computation.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+  var dotIndex = -1
+  var json
```
fpagnoux: Same than `dotIndex `. Why not initiate this variable inside the function ?

src/server/index.js
```diff
@@ -76,11 +77,16 @@ Promise.all([fetchParameters(), fetchVariables()])
       assoc('itemType', 'variable'),
       addNormalizedDescription(variablesResponse.data),
     )
+
+    console.log("Fetching text translations...")
```
MattiSG: Don't commit `console` statements.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
+import path from "path"
+import {readdir} from "fs"
+
+import {addLocaleData} from "react-intl"
+import fr from "react-intl/locale-data/fr"
+import en from "react-intl/locale-data/en"
+
+const DEFAULT_LANGUAGE = 'fr'
+
+export function loadTranslations(langDir){
```
MattiSG: Missing space between `)` and `{`.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+  addLocaleData(...fr)
+  addLocaleData(...en)
+
+  var messages = new Map()
+  var dotIndex = -1
+  var json
+  readdir(langDir, (err, files) => { 
+    if(err){
```
MattiSG: Spaces.

fpagnoux:
There is an error:
```
Uncaught TypeError: e.messages.get is not a function
    at a (client.jsx:43)
    at Object.<anonymous> (client.jsx:56)
    at e (bootstrap dc2bd8f…:19)
    at bootstrap dc2bd8f…:39
    at bootstrap dc2bd8f…:39
````
This is what make the tests fail.

File: src/server/lang.js [hunk 9]
```diff
@@ -0,0 +1,43 @@
...
+      console.log("Unable to load translation files.", err)
+    }
+
+    files.forEach(file => {
+      dotIndex = file.indexOf('.')
+      json = path.resolve(langDir, file)
+      messages.set(file.substring(0, dotIndex), require(json))
```
MattiSG: Use [`path.basename`](https://nodejs.org/api/path.html#path_path_basename_path_ext)… or `.replace('.json', '')`.

fpagnoux:
I tried to push a fix

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
+import path from "path"
+import {readdir} from "fs"
+
+import {addLocaleData} from "react-intl"
+import fr from "react-intl/locale-data/fr"
+import en from "react-intl/locale-data/en"
```
MattiSG: Shouldn't these be dynamically required from `loadTranslations`? Otherwise, I don't see how you can load a full directory by iterating over whatever's defined in it.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+export function loadTranslations(langDir){
+  addLocaleData(...fr)
+  addLocaleData(...en)
+
+  var messages = new Map()
+  var dotIndex = -1
+  var json
+  readdir(langDir, (err, files) => { 
+    if(err){
+      console.log("Unable to load translation files.", err)
```
MattiSG: 1. This is not a `log` but an `error`.
2. Can you actually continue? Are you sure you shouldn't `throw`?
3. Improve the error message by explaining which `langDir` you were trying to load.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    files.forEach(file => {
+      dotIndex = file.indexOf('.')
+      json = path.resolve(langDir, file)
```
MattiSG: This is not actually a `json`, this is the path to a set of locale translations.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+      messages.set(file.substring(0, dotIndex), require(json))
+    })
+  })
+  return messages
```
MattiSG: > I would recommend extracting to function the loading of one specific locale based on its name, and simply iterating over `readdir` results.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function getLocale(acceptLanguage, messages){
+  var locale = acceptLanguage ? acceptLanguage.substring(0, 2) : DEFAULT_LANGUAGE //ex: en-US;q=0.4,fr-FR;q=0.2 > en
+
+  if(!messages.has(locale)){
+    console.log('Unsupported locale "' + locale + '". Switching to default language: ' + DEFAULT_LANGUAGE)
```
MattiSG: I don't see the point of logging this in the backend.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+      dotIndex = file.indexOf('.')
+      json = path.resolve(langDir, file)
+      messages.set(file.substring(0, dotIndex), require(json))
+    })
+  })
+  return messages
+}
+
+export function getLocale(acceptLanguage, messages){
+  var locale = acceptLanguage ? acceptLanguage.substring(0, 2) : DEFAULT_LANGUAGE //ex: en-US;q=0.4,fr-FR;q=0.2 > en
```
MattiSG: This is not correct: you only accept a match on the first user preference. Do not reinvent parsing at all and add more functionality by using a package such as https://github.com/tinganho/node-accept-language.

src/server/lang.js
```diff
@@ -0,0 +1,43 @@
...
+
+  if(!messages.has(locale)){
+    console.log('Unsupported locale "' + locale + '". Switching to default language: ' + DEFAULT_LANGUAGE)
+    locale = DEFAULT_LANGUAGE
+  }
+  return locale
+}
+
+export function getLocaleMessages(locale, messages){
+  return messages.get(locale)
```
MattiSG: What is the added value of such a wrapper function? Why not simply expose `messages`?

tests/AppFixture.js
```diff
@@ -1,2 +1,3 @@
-baseline = /\d+ variables et \d+ paramètres référencés/
+baseline = /OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies./
```
MattiSG: Increase test resilience and readability by matching only a significant subset of the text.

tests/AppFixture.js
```diff
@@ -1,2 +1,3 @@
...
+stats = /\d+,\d+ variables and \d+,\d+ parameters referenced/
```
MattiSG: This assumes that there will always be over 1,000 variables and parameters. Rather use `[0-9,]+`.

src/components/app.jsx
```diff
@@ -65,8 +80,12 @@ const App = React.createClass({
...
+                    <br/>
+                    <small id="stats">
+                      <FormattedMessage {...messages.stats} values={{nbv: Object.keys(variables).length, nbp: Object.keys(parameters).length}}/>
```
MattiSG: Why use cryptic 3-letters identifiers? `variablesCount` and `parametersCount` would be much more readable 😉

src/server/index.js
```diff
@@ -76,11 +77,16 @@ Promise.all([fetchParameters(), fetchVariables()])
...
+    const messages = loadTranslations(path.resolve(__dirname, "../assets/lang/"))
```
MattiSG: > Are you sure you need to explicitly `path.resolve(__dirname, …)`?

src/components/app.jsx
```diff
@@ -3,12 +3,27 @@ import url from "url"
...
 import * as AppPropTypes from "../app-prop-types"
 import config from "../config"
 import {findParametersAndVariables} from "../search"
 
 
+const messages = defineMessages({
+    header: {
+        id: 'header',
+        defaultMessage: 'OpenFisca rend le droit calculable. '
+          + 'Il s’agit d’un moteur de calcul libre et ouvert qui permet de modéliser le système socio-fiscal de manière collaborative et transparente. '
```
sandcha: Removed since the messages are now read on server side.

src/components/app.jsx
sandcha: <span> added instead as there is already a <p> as parent.

src/server/index.js
sandcha: Removed but there are 10 older occurrences of `console.log` in the same file. In the future, what should we use instead ?

src/server/index.js
sandcha: Updated to `path.join`.

src/server/lang.js
sandcha: var removed.

src/server/lang.js
sandcha: To separate language management from other work.

fpagnoux:
Rebasing and merging

MattiSG:
Has this been cross-browser tested ?

sandcha:
Yes, cross-browser tested.

maukoquiroga:
How did you cross-browser tested it? The way to set ```accept-language``` changes by the driver chosen, and can't be set in IE.

Issue #98, opened 2017-06-22
Link variable names to their description page


MattiSG:
In the source code of a variable, variable names that are documented in the legislation explorer are clickable and link to their description page.

PR #101, opened 2017-06-28
Clickable variable names 98

sandcha:
Connected to #98 

In a variable page, replace any reference to an OpenFisca variable by a link to its page.

### DIFF
--- a/src/components/variable.jsx
+++ b/src/components/variable.jsx
@@ -1,7 +1,7 @@
 import DocumentTitle from "react-document-title"
 import { FormattedDate } from "react-intl"
 import React, { PropTypes } from "react"
-import {routerShape, locationShape, Link} from "react-router"
+import {Link} from "react-router"
 import { keys } from "ramda"
 
 import config from "../config"
--- a/src/components/variable.jsx
+++ b/src/components/variable.jsx
@@ -11,10 +11,6 @@ import getDayBefore from "../periods"
 
 
 const Variable = React.createClass({
-  contextTypes: {
-    router: routerShape.isRequired,
-    searchQuery: PropTypes.string.isRequired,
-  },
   propTypes: {
     countryPackageName: PropTypes.string.isRequired,
     countryPackageVersion: PropTypes.string.isRequired,
@@ -24,29 +20,6 @@ const Variable = React.createClass({
   },
   getTodayInstant() {
     return new Date().toJSON().slice(0, 10)
-  },
-  navigateProgramatically(e) {
-    const searchQuery = ""
-    this.setState({variable: searchQuery})
-    this.context.setSearchQuery(searchQuery)
-    this.context.router.push({
-      query: {q: searchQuery},
-    })
-  },
-  componentDidMount() {
-    this._isMounted = true
-    const {router} = this.context
-    this.unregisterRouterListen = router.listen(this.locationHasChanged)
-  },
-  locationHasChanged(location) {
-    const {router} = this.context
-    const oldLocation = this.props.location
-    // Check that the new location stays on the Home page, to avoid overwriting searchQuery in App state.
-    if (this._isMounted && oldLocation && router.isActive(oldLocation)) {
-      const searchQuery = location.query.q || ""
-      this.context.setSearchQuery(searchQuery)
-      this.setState({variable: variable.data, location: location, waitingForResponse: false})
-    }
   },
   render() {
     const {variable} = this.props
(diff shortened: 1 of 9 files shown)


src/components/variable.jsx
```diff
@@ -107,14 +135,29 @@ const Variable = React.createClass({
...
+  link(substring, quote) {
+    if (!substring.includes(" ") && this.props.variables[substring]) {
+      return <Link key={substring} to={substring} onClick={this.navigateProgramatically}>{quote + substring + quote}</Link>
+    } 
+    return quote + substring + quote
+  },
+  // Change every OpenFisca variable in the formula by a link to the variable page:
+  renderLinkedFormulaVariables(formula) {
+    return formula.split("'").map((substring, index) => {
+      return (index % 2 != 0) ?
```
fpagnoux: What if there is a single `'` in the comments of the formula ?

For instance:
```py
def formula(famille, period):
    # This ' will break the algorithm
    x = famille('x', period)
    y = famille('y', period)
    return x + y
```

src/server/webpack-dev-server.js
```diff
@@ -14,6 +14,7 @@ new WebpackDevServer(webpack(config), {
   noInfo: true,
   publicPath: config.output.publicPath,
   stats: { colors: true },
+  quiet: true,
```
fpagnoux: > 🎉

src/components/variable.jsx
```diff
@@ -107,14 +135,29 @@ const Variable = React.createClass({
...
+                <pre><code>{this.renderLinkedFormulaVariables(formulas[date].content)}</code></pre>
               </div>
             )
           })
         }
       </div>
     )
   },
+  link(substring, quote) {
+    if (!substring.includes(" ") && this.props.variables[substring]) {
```
fpagnoux: > !substring.includes(" ")

Seems like a fair optimization: you only keep single words 👍 . Maybe a comment to explain why ?

fpagnoux: According to the functions naming, I'd expect `renderLinkedFormulaVariables` to check that the substrings are valid variables, and `link` to just build a `Link`.

src/components/variable.jsx
```diff
@@ -20,6 +25,29 @@ const Variable = React.createClass({
   getTodayInstant() {
     return new Date().toJSON().slice(0, 10)
   },
+  navigateProgramatically(e) {
```
fpagnoux: This seems complex. It seems that we are re-implementing routing. 

What problem that you encountered brought you to write this ? Did you get help from anyone who knows react well ?

fpagnoux:
Also, an integration test would be welcome :).

src/components/variable.jsx
fpagnoux: Maybe check there: https://github.com/ReactTraining/react-router/issues/4975

fpagnoux: No, this seems to be the problem: https://stackoverflow.com/questions/32261441/component-does-not-remount-when-route-parameters-change

src/components/variable.jsx
```diff
@@ -107,14 +135,29 @@ const Variable = React.createClass({
...
   },
+  link(substring, quote) {
+    if (!substring.includes(" ") && this.props.variables[substring]) {
+      return <Link key={substring} to={substring} onClick={this.navigateProgramatically}>{quote + substring + quote}</Link>
+    } 
+    return quote + substring + quote
+  },
+  // Change every OpenFisca variable in the formula by a link to the variable page:
+  renderLinkedFormulaVariables(formula) {
+    return formula.split("'").map((substring, index) => {
```
sandcha: Ok but we might have to loop on the splits after the first map.

src/components/variable.jsx
```diff
@@ -107,14 +108,44 @@ const Variable = React.createClass({
...
+    })
+  },
+  //Change every OpenFisca variable in the formula by a link to the variable page:
+  renderLinkedFormulaVariables(formula) {
+    //Split on double quotes first (preventing collision with Link):
+    var splits = this.linkCodeSplits(formula.split('"'), '"') 
+    for (var s of splits) {
+      s = this.linkCodeSplits(s.split("'"), "'")
+    }
+    return s
```
fpagnoux: That seems to be only returning the last split... I doubt this works in a variable with several variables calls surrounded by `""`

src/components/variable.jsx
```diff
@@ -107,14 +108,44 @@ const Variable = React.createClass({
...
+    return (!substring.includes(" ") && this.props.variables[substring])
+  },
+  linkCodeSplits(splits, separator) {
+    var previousIsLink = false
+    return splits.map((substring, index) => {
+      if (this.isVariable(substring)) {
+        substring = this.link(substring)
+        previousIsLink = true
+      } else {
+        substring = previousIsLink ? separator + substring : substring  //No jsx and separator concatenation.
```
fpagnoux: I don't really understand... Why do we need to add a separator only if the previous item was a `Link` ?

Issue #84, opened 2017-05-18
Handle multiple words research


fpagnoux:
As a legislation explorer user typing "salaire base" in the search field, 
I get the variable "salaire_de_base" in my results

MattiSG:
Should supersede #55 if parameters are properly named.

MattiSG:
1. Select only variables where name contains each query words in either variable name or description.
2. Group variables according to query word match count in variable name and order them in decreasing order.
3. Group variables according to query word match count in variable description and order them in decreasing order.
4. Group each subset according to query word match index sum.
5. Sort each sub-subset according to alphabetical order.
6. Concatenate all sub-subsets and present them as ordered results list.

## Examples

`salaire base` →
- `salaire_de_base` _Salaire de base, en général appelé salaire brut, la 1ère ligne sur la fiche de paie_
- `cotsoc.indemnite_fin_contrat.taux` _Pourcentage du salaire de base_

sandcha:
1 => Every result must cover all query words ; remove results that doesn't.

Besides, query word repetition in a variable/parameter name and description has no influence ; do not change the order/weight of such results.

PR #102, opened 2017-07-03
Search multiple words

sandcha:
Connected to #84 

Manage multiple words search.
First words have more importance than next ones.
Multiple matching in name has priority.
And it looks for matching in name and in description but doesn't look for word repetition in name and in description.


fpagnoux:
If I type `salaire base`, I get results I should not have:

image

fpagnoux:
Sorry I had to rewrite some of the code , but the demo is coming 😱 .

As I can't review what is not here anymore, some feedback:

- In `JS`, as in duck-typed languages in general, you are quite flexibles on types. For instance  `1 + true == 2`. So  `counter += condition ? 1 : 0` can be easily replaced by `counter += condition`.

- When you want to iterate on the items of an array, the fancy and brand new `for (const item of array) {` is a nicer than `for (const item in array) {`. The latter will iterate on the keys `i`, and you then have to get the value with `array[i]`, which is very heavy. If you can't use `ES6` on a other project, `array.forEach(...)` is still a better alternative than iterating on keys.

In general, it's more readable to avoid working with the indexes, unless you _really_ have to (and you rarely have).

- For this kind of code, automated testing is crucial. The algorithm is a little tricky, and you don't want to manually check that you still respect _all_ the spec _each time_ you change something. 

Otherwise, storing the scores as properties was really clever, and the sorting was well done 👍 .

fpagnoux:
I'm not able to review this, as I edited the code.
@MattiSG do you have time ?
Other @sandcha or @Anna-Livia ?

.gitignore
```diff
@@ -4,5 +4,6 @@
 /public/
 /tags
 /webpack-assets.json
+.tmp/mocha-webpack/
```
fpagnoux: We don't use `mocha-webpack` anymore ;)

tests/.gitignore
```diff
@@ -0,0 +1 @@
+/npm-debug.log
```
fpagnoux: Why not putting this in the root `.gitignore` ?

sandcha:
Ok with your code changes as well.

fpagnoux:
Ok, rebasing and merging 🚀

PR #106, opened 2017-07-07
Improve readability for Variables

Anna-Livia:
Fixes #79 

This PR : 
- translates the Variable Page
- formats the variable's metadata so they are more readable
- adds edit links to each formula

---------

EDIT by @fpagnoux:

Depends on https://github.com/openfisca/openfisca-doc/pull/80

PR #113, opened 2017-08-07
Add links to header rather than switch buttons

### LINKED ISSUE #89: Do not give the impression that legislation / API switch is a tab toggle
**If I:**

- Search for “smic”.
- Click on “smic_proratise”.
- Click on “API”.
- Click on “Legislation”.

**Then:**

- I am redirected to the homepage.

**While I expected:**

- To be redirected to the “smic_proratise” page, and able to go back to my “smic” research.

Anna-Livia:
Fix #89.

This PR replaces the "Legislation" and "API" button from the header with links inside the header text.

The links are bold to increase discoverability.


src/assets/lang/en.json
```diff
@@ -1,6 +1,10 @@
 {
   "header": "OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
-  "stats": "{variablesCount, number} variables and {parametersCount, number} parameters referenced to modelize the french tax-benefit system.",
+  "learnMore":"Learn more.",
+  "stats": "{explorerLink} that modelize the french tax-benefit system.",
```
MattiSG: modelize → model

MattiSG: french → French

MattiSG: tax-benefit → tax and benefit

src/assets/lang/en.json
```diff
@@ -1,6 +1,10 @@
 {
   "header": "OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
-  "stats": "{variablesCount, number} variables and {parametersCount, number} parameters referenced to modelize the french tax-benefit system.",
+  "learnMore":"Learn more.",
```
MattiSG: Missing space.

src/assets/lang/en.json
```diff
@@ -1,6 +1,10 @@
...
+  "stats": "{explorerLink} that modelize the french tax-benefit system.",
+  "explorerText":"Explore the {variablesCount, number} variables and {parametersCount, number} parameters",
+  "api":"{apiLink} that lets you compute them from any software.",
```
MattiSG: any software → any other piece of software / other software / any app ? 🤔

src/assets/lang/en.json
```diff
@@ -1,6 +1,10 @@
...
+  "apiText":"Use the API",
```
MattiSG: Missing space (repeated above).

src/components/app.jsx
```diff
@@ -80,23 +70,49 @@ const App = React.createClass({
                 <div className="col-md-9">
                   <p id="baseline">
                     <span className="message">
-                      <FormattedMessage {...messages.header} />
+                      <FormattedMessage id = "header"/>
```
MattiSG: No spaces around `=` on attributes.

src/components/app.jsx
```diff
@@ -80,23 +70,49 @@ const App = React.createClass({
...
+                      <small>
+                        <ExternalLink href = "https://www.openfisca.fr" target = "_blank">
+                          <strong><FormattedMessage id = "learnMore"/></strong>
```
MattiSG: > Using CSS on `header a` would be more welcome than using `<strong>`.

MattiSG:
Nice!  :)

src/assets/lang/en.json
Anna-Livia: 👍

src/assets/lang/en.json
Anna-Livia: 👍

src/assets/lang/en.json
Anna-Livia: "any app" :raised_hands:

src/assets/lang/en.json
Anna-Livia: 🤔 Je ne vois pas ...

src/components/app.jsx
Anna-Livia: J'ai suivi la reco de la page `variables`---> https://github.com/openfisca/legislation-explorer/pull/106#discussion_r126416085

src/assets/lang/en.json
MattiSG: In JSON, the usual convention is to have a space after colons in property name/value pairs, like so:

```
"apiText": "Use the API"
          ^
```

🙂

src/components/app.jsx
MattiSG: It is not the same thing to introduce a specific class and to target already existing markup 🙂
Also, I'm not sure I agree with @fpagnoux on this specific case.

src/assets/lang/en.json
```diff
@@ -1,6 +1,10 @@
 {
-  "header": "OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
-  "stats": "{variablesCount, number} variables and {parametersCount, number} parameters referenced to modelize the french tax-benefit system.",
+  "header":"OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
+  "learnMore":" Learn more.",
```
MattiSG: Why is there a leading space inside of the message? This usually makes it harder to compose messages when they include leading or trailing space.
My previous comment on a missing space was after the JSON colon 🙂

src/components/app.jsx
```diff
@@ -80,23 +70,49 @@ const App = React.createClass({
                 <div className="col-md-9">
                   <p id="baseline">
                     <span className="message">
-                      <FormattedMessage {...messages.header} />
+                      <FormattedMessage id="header"/>
+                      <small>
+                        <ExternalLink href = "https://www.openfisca.fr" target = "_blank">
```
MattiSG: No space around `=` on attributes. This is not an imperative allocation, but an XML-style declaration. The fact that JSX supports spaces should not make it harder for a human eye used to XML to read this code.
Please, this is repeated on several occurrences and was already requested for change in the previous review.

fpagnoux: It's probably worth adding [this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-equals-spacing.md) to the [linter](https://github.com/openfisca/legislation-explorer/blob/master/.eslintrc). 🙂

src/components/app.jsx
fpagnoux: Was `<span class="variableMetadataHighlight">` better than `<strong>` ? If so, more explanations would be welcome.

src/components/app.jsx
Anna-Livia: Done :)

src/assets/lang/en.json
Anna-Livia: I hope I fixed it this time 🤔

src/components/app.jsx
MattiSG: @fpagnoux It could be as it is more semantic (I would argue for `class="metadata"` which is less specific). I am lacking context on that case, though. But in the current case, where there is a native selector that does make sense (we want to increase the font weight of the `header a` because links show up in a lighter colour that has to be compensated only because they are in the header, which has a lighter font-weight), it would be better to use CSS than markup. There is nothing semantic in the `<strong>`: the meaning is already, and better, carried by the `<a>`.

src/assets/lang/en.json
MattiSG: No :wink:

Trailing & leading space in message strings is not a good practice. You should leave it to the message composer (the part that assembles the different strings in the interface) to decide where to put spaces. For example, if you were to enclose the message in some HTML tags, leading or trailing spaces would not be collapsed together.

Once again: 

> My previous comment on a missing space was after the JSON colon

The expectation is:

```diff
-  "learnMore":" Learn more.",
+  "learnMore": "Learn more.",
               ^^
``` 

I'll handle these changes.

src/assets/lang/en.json
```diff
@@ -1,7 +1,11 @@
 {
-  "header": "OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies.",
-  "stats": "{variablesCount, number} variables and {parametersCount, number} parameters referenced to modelize the french tax-benefit system.",
-  "entityParagraph" : "This {variableLink} applies to the {entityLink} ",
+  "header":"OpenFisca makes law computable. It is a free and open-source computation engine that allows modelling tax and benefit systems in a collaborative and transparent way. This software is used by researchers and public agencies. ",
+  "learnMore":"Learn more.",
```
MattiSG: No :wink:

Trailing & leading space in message strings is not a good practice. You should leave it to the message composer (the part that assembles the different strings in the interface) to decide where to put spaces. For example, if you were to enclose the message in some HTML tags, leading or trailing spaces would not be collapsed together.

Once again: 

> My previous comment on a missing space was after the JSON colon

The expectation is:

```diff
-  "learnMore":" Learn more.",
+  "learnMore": "Learn more.",
               ^^
``` 

I'll handle these changes.

PR #116, opened 2017-08-16
Add a title for parameters pages

### LINKED ISSUE #115: Add a title for parameters pages
On a [parameter page](https://legislation.openfisca.fr/prestations.al_plaf_acc.plafond_pour_accession_a_la_propriete_zone_3.menage_ou_isole_avec_4_enfants): 

- I can't see the path towards the parameter. I need this path to use the parameter in my formula code.
- If there is no description for the parameter, the page title is `Aucune description`.

sandcha:
Fixes #115 

For a parameter:
* Update the page title to show the parameter id instead of its description
* Add the parameter id in the page and split it when it's long (but keep it in one selectable line)


src/assets/style.css
```diff
@@ -1,12 +1,17 @@
+header {
+  word-wrap:break-word;
```
MattiSG: Missing space after colon.

src/components/parameter.jsx
```diff
@@ -19,12 +19,14 @@ const Parameter = React.createClass({
     const {parameter} = this.props
     const description = parameter.description || "Aucune description"
     const isScale = (! parameter.values)
+    const multilineId = parameter.id.replace(/[.]/g, '\u200b.')
```
MattiSG: > Escape the dot (`\.`) rather than using a single-entity character set (`[.]`), this is less costly.

MattiSG: > If `&shy;` works, prefer it over `\u200b`, which is less readable.

src/assets/style.css
MattiSG: In which cases is this necessary? I thought we added the `&shy;` exactly to avoid breaking mid-word. 🤔

sandcha: Removed.

src/components/parameter.jsx
sandcha: `&shy;` appears in the string. 😕 `\u200b` kept but comment added.

MattiSG:
capture d ecran 2017-08-18 a 17 22 01
capture d ecran 2017-08-18 a 17 22 11

PR #130, opened 2017-11-23
Display new Enums in variable pages

Anna-Livia:
Connected to openfisca/openfisca-france#831

A enum variable page now displays a keyword/description table to indicate to users which keyword they should choose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,11 @@ To maintain the legislation at its latest version, you can check the [auto-updat
 
 Install [Selenium standalone server](http://www.seleniumhq.org/download/).
 
-Install Chrome WebDriver : The `tests/config.js` in this package is pre-configured to use Chrome WebDriver. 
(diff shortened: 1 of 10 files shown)


sandcha:
> Environment updated ; Chrome 62.0.3202.94 needed a chromedriver > 2.30 (aka 2.33). 
One integration test error that seems quick to fix: `Found text "celibataire" and no value in VariableComponent.firstPossibleValue, which does not match "/marie/".`

Besides, is the following line still true `It does not have a formula and therefore is not computable. It can only be used as an input for a computation.`? @fpagnoux @guillett Could an enum value be deduced during computation?

guillett:
Yes, [zone_apl](https://fr.openfisca.org/api/v18/variable/zone_apl) is such an Enum.

maukoquiroga:
Rebase + fixup

PR #131, opened 2017-12-09
Fix integration tests

maukoquiroga:
Integration test are failing because one test is using enums but they're not yet available.


Anna-Livia:
If everything goes according to plan, the test should show `marie` as appearing as the default value.
Changing the test back to `0`means the Enum changes would not have been implemented.

maukoquiroga:
Roger that 👍

PR #134, opened 2018-01-17
Switch back circle CI to production like configuration

fpagnoux:


maukoquiroga:
Isn't it enough to just comment out

```
-    - npm run dev:
-        background: true
-        parallel: true
```

PR #140, opened 2018-04-09
Set default OpenFisca France API version to v21

sandcha:
Connected to openfisca/openfisca-ops#42.
Move default France API version from v20 to v21.

### DIFF
--- a/config/production.france.js
+++ b/config/production.france.js
@@ -4,7 +4,7 @@
 import winston from "winston"
 
 
-const apiBaseUrl = process.env.API_URL || `https://fr.openfisca.org/api/v20`,
+const apiBaseUrl = process.env.API_URL || `https://fr.openfisca.org/api/v21`,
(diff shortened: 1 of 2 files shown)


config/production.france.js
```diff
@@ -4,7 +4,7 @@
 import winston from "winston"
 
 
-const apiBaseUrl = process.env.API_URL || `https://fr.openfisca.org/api/v20`,
+const apiBaseUrl = process.env.API_URL || `https://fr.openfisca.org/api/v21`,
```
maukoquiroga: You can set that directly in your [service definition](https://github.com/openfisca/openfisca-ops/blob/3aead030dbaa0963430bcc9cb35c10ad21f48851/fr.openfisca.org/api/v21/openfisca-web-api-fr21.service)

sandcha: 👍 Not exactly in this PR perimeter but I agree, this configuration could be extracted from `legislation explorer`

PR #144, opened 2018-05-09
Use international website in description

MattiSG:
We want “learn more” here to link to openfisca.org and let the target language be set dynamically rather than hardcoding French.

screen shot 2018-05-09 at 15 21 54


MattiSG:
Superseded by #149.

PR #161, opened 2018-06-09
Use prop-types as React.PropTypes is deprecated

maukoquiroga:
These changes are required if we want to upgrade our app's React version.


src/components/app.jsx
```diff
@@ -1,5 +1,6 @@
 import DocumentTitle from "react-document-title"
-import React, {PropTypes} from "react"
+import React from "react"
+import PropTypes from 'prop-types'
```
MattiSG: Use double quotes as in other import statements.

MattiSG:
According to https://stackoverflow.com/questions/43289348 this is probably because some dependency is still using React.PropTypes. I feel that this PR would be complete only if the warning is removed.

MattiSG:
I understand, what I'm telling you is that the app dependencies can't currently be upgraded 😉

maukoquiroga:
> Not blocking, but not approving until understanding of the origin of the warning has been obtained.

Haven't found the origin yet.

maukoquiroga:
`react-router` !

maukoquiroga:
Oh! Ok I'll take a look.

maukoquiroga:
`swagger-ui`!

Anna-Livia:
@maukoquiroga FYI, there is a linter alert 
It would be good to change it before rebase & merge :)
```
/Users/octo-ango/Desktop/Anna-Livia/openfisca/originals/legislation-explorer/src/history.js
  6:3  error  Newline required at end of file but not found  eol-last
```

maukoquiroga:
Hi @Anna-Livia, I can't find that file in the the repository.

PR #173, opened 2018-08-16
Take .env into account when building webpack for prod 

fpagnoux:
See https://github.com/openfisca/legislation-explorer/pull/171 for issue description

### DIFF
--- a/webpack.config.prod.babel.js
+++ b/webpack.config.prod.babel.js
@@ -5,8 +5,7 @@ import webpack from 'webpack'
 
 import writeAssets from './src/server/write-assets'
 
-const dotenv = require('dotenv').config('./.env').parsed
-process.env = Object.assign({}, dotenv, process.env)
+require('dotenv').config()
 
 const config = require('./src/config').default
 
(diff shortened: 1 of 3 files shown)


webpack.config.prod.babel.js
```diff
@@ -3,9 +3,11 @@ import path from 'path'
 import CopyWebpackPlugin from 'copy-webpack-plugin'
 import webpack from 'webpack'
 
-import config from './src/config'
 import writeAssets from './src/server/write-assets'
 
+require('dotenv').config()
```
fpagnoux: > This will redefine  `process.env` to take into account the `.env` file

webpack.config.prod.babel.js
```diff
@@ -3,9 +3,11 @@ import path from 'path'
...
+
+const config = require('./src/config').default
```
fpagnoux: > The `config` module code needs to be executed _after_ `process.env` has been redefined, so we can't `import` it and need to `require` it.

fpagnoux:
Another approach that seems to work is to change `npm run build` from `webpack --config ./webpack.config.prod.babel.js` to `node --require dotenv/config node_modules/webpack/bin/webpack.js --config ./webpack.config.prod.babel.js`

A little ugly  but more consistent with `npm run dev` and `npm start`.

Reviewers, let me know if you have a preference :)

Anna-Livia:
I like `node --require dotenv/config node_modules/webpack/bin/webpack.js --config ./webpack.config.prod.babel.js`  better because of the constistency with the other scripts.

PR #186, opened 2018-08-20
Re-add continuous deployment

fpagnoux:
Connected to openfisca/openfisca-ops#60
Depends on https://github.com/openfisca/openfisca-ops/pull/61

It works: https://circleci.com/gh/openfisca/legislation-explorer/714

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,7 @@ jobs:
       WATAI_TESTS: tests/integration
       SAUCE_USERNAME: openfisca
       # SAUCE_ACCESS_KEY is defined in the CircleCI web UI
+      # DEPLOY_SSH is defined in the CircleCI web UI
       PORT: 2030 # Legislation Explorer ExpressJS server port
 
     steps:
@@ -74,7 +75,7 @@ jobs:
     steps:
       - run:
           name: Deploy app
-          command: ssh -o StrictHostKeyChecking=no deploy-legislation-explorer@fr.openfisca.org
+          command: if [ -n "$DEPLOY_SSH" ]; then ssh -o StrictHostKeyChecking=no $DEPLOY_SSH; fi
 workflows:
   version: 2
   build:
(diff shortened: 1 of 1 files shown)


fpagnoux:
> Is there a way to make this config?

Hmm, CircleCI config is pretty limited. The only way I could see to have a different behavior in the fork would be to use a environment variable, something like:

```sh
if [ -n "$DEPLOY" ]; then ssh -o StrictHostKeyChecking=no deploy-legislation-explorer@fr.openfisca.org; fi
```

However, I'm wondering if this would really be a good way to scale continuous deployment for a multi-instance app like the legislation explorer. One solution I could imagine would be, on master merges, to trigger _n_ deployments to deploy the legislation explorer for _n_ countries:

```YAML
  deploy:
    docker:
      - image: circleci/node:8-jessie-browsers
    steps:
      - run:
          name: Deploy France app
          command: ssh -o StrictHostKeyChecking=no deploy-legislation-explorer@fr.openfisca.org
      - run:
          name: Deploy New Zealand app
          command: ssh -o StrictHostKeyChecking=no deploy-legislation-explorer@nz.openfisca.org
      - run:
          name: Deploy Tunisia app
          command: ssh -o StrictHostKeyChecking=no deploy-legislation-explorer@tn.openfisca.org
```

I have more trouble imagining what workflow could work with forks... Do you have one in mind?

To make sure I understand your need correctly, what is the benefit of using a fork, instead of directly contributing on this repository? 

Thanks in advance 🙂

Br3nda:
Can you put this bit in an env variable in circle ci:
`deploy-legislation-explorer@fr.openfisca.org`
Then only deploy if it's set.

fpagnoux:
So something like:
```sh
if [ -n "$DEPLOY_SSH" ]; then ssh -o StrictHostKeyChecking=no $DEPLOY_SSH; fi
```

Seems a little annoying to scale if we had several servers to deploy to... Ok for now as France seems to be, as of today, the only country interested in continuous deployment.

fpagnoux:
Ok, makes sense, thanks for the explanation!

PR #187, opened 2018-08-20
Display an error message when the API doesn't respond

### LINKED ISSUE #136: Frozen page on api internal server error
When I search for `https://legislation.openfisca.fr/paje_cmg`, the page stays on `Chargement des valeurs…`.

We could give more information to the user as on this peculiar context, an API call to this variable `https://fr.openfisca.org/api/v20/variable/paje_cmg` gives: 
```
{
  "error": "Internal server error: 'ascii' codec can't decode byte 0xc3 in position 112: ordinal not in range(128)"
}
```

fpagnoux:
Fixes #136 

Before:
image

After:

image

### DIFF
--- a/src/server/webpack-dev-server.js
+++ b/src/server/webpack-dev-server.js
@@ -15,6 +15,9 @@ new WebpackDevServer(webpack(webpackDevConfig), {
   publicPath: webpackDevConfig.output.publicPath,
   stats: { colors: true },
   quiet: true,
+  headers: {
+    'Access-Control-Allow-Origin': '*',
+  },
 }).listen(publicURL.port, publicURL.hostname, function (err) {
   if (err)
     return console.error(err)
--- a/src/webservices.js
+++ b/src/webservices.js
@@ -5,19 +5,15 @@ import config from './config'
 
 async function fetchJson(url, options) {
   const response = await fetch(url, options)
+  if (response.status < 200 || response.status >= 300) {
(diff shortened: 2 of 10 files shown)


src/server/webpack-dev-server.js
```diff
@@ -15,6 +15,9 @@ new WebpackDevServer(webpack(webpackDevConfig), {
   publicPath: webpackDevConfig.output.publicPath,
   stats: { colors: true },
   quiet: true,
+  headers: {
+    'Access-Control-Allow-Origin': '*',
```
fpagnoux: > Avoid CORS error when using an `.env` file locally

webpack.config.dev.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
   devtool: 'source-map', // Original code
   entry: {
     'main': [
+      'babel-polyfill',
```
fpagnoux: > To be able to use `async/await`

src/webservices.js
```diff
@@ -3,23 +3,17 @@ import fetch from 'isomorphic-fetch'
...
-          }
-        }
-        if (data.error) {
-          throw new Error(JSON.stringify(data.error))
-        }
-        throw new Error(JSON.stringify({error: 'Unexpected return code ' + response.status}))
-      })
-  )
+async function fetchJson(url, options) {
+  const response = await fetch(url, options)
```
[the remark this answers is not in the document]
fpagnoux: Added it 🙂

PR #188, opened 2018-08-21
Upgrade React

fpagnoux:
Depends on #187 

_Note: This PR is opened against `error-message` for diff readability, but should be merged on `master`. Target branch should be changed once #187  is merged._


fpagnoux:
Arg, seems like the integration tests repeatedly fail for Android 😞 

No ready to merge yet.

fpagnoux:
I'm reducing the scope of the upgrades as for some reason I can't get the Android integration tests to pass after upgrading Webpack 😕.

This is now ready for review!

src/components/pages/home.jsx
```diff
@@ -11,35 +11,37 @@ import config from '../../config'
...
     countryPackageName: PropTypes.string.isRequired,
     countryPackageVersion: PropTypes.string.isRequired,
     intl: intlShape,
     location: locationShape.isRequired,
     parameters: PropTypes.objectOf(parameterShape).isRequired,
     variables: PropTypes.objectOf(variableShape).isRequired,
-  },
+  };
+
+  state = {inputValue: ''};
```
sandcha: `this.state`?

src/components/pages/parameter-or-variable.jsx
```diff
@@ -10,55 +10,65 @@ import { searchInputId } from './home'
...
     parameters: PropTypes.objectOf(parameterShape).isRequired,
     variables: PropTypes.objectOf(variableShape).isRequired,
-  },
-  getInitialState() {
-    return {variable: null, parameter: null, waitingForResponse: true}
-  },
-  fetchPageContent(name) {
+  };
+
+  state = {variable: null, parameter: null, waitingForResponse: true};
```
sandcha: `this.state`?

src/components/pages/parameter-or-variable.jsx
```diff
@@ -10,55 +10,65 @@ import { searchInputId } from './home'
...
       this.handleNotFound()
     }
-  },
-  componentWillReceiveProps(nextProps) {
-    this.fetchPageContent(nextProps.params.name)
-  },
+  };
+
+  componentDidUpdate(prevProps) {
+    if (this.props.params.name != prevProps.params.name) {
```
sandcha: 🙂

src/components/pages/searchbar.jsx
```diff
@@ -4,32 +4,32 @@ import {FormattedMessage, injectIntl, intlShape} from 'react-intl'
...
+class SearchBarComponent extends React.Component {
+  static contextTypes = {
     router: routerShape.isRequired,
-  },
+  };
 
-  getInitialState() {
-    return {inputValue: this.props.initialValue}
-  },
+  state = {inputValue: this.props.initialValue};
```
sandcha: Same note: `this.state`?

src/components/variable.jsx
```diff
@@ -144,18 +144,18 @@ const Variable = React.createClass({
...
-                      </td>
-                      <td>
-                        <samp>{variable.possibleValues[keyword]}</samp>
-                      </td>
-                    </tr>
-                  )
-                })
+                  Object.keys(variable.possibleValues).map((keyword) => {
+                    return (
+                      <tr key={keyword}>
```
sandcha: Is this `key` used somewhere or is it here to bring similarity between variables and parameters? 🤔

src/components/pages/home.jsx
fpagnoux: That wouldn't work:

Here, we are defining the initial value of a class attribute. We are not within a method, and using `this` would trigger a syntax error. 

We would use `this` if we initialized the state from a constructor, which is [another way](https://stackoverflow.com/questions/46714799/how-to-set-initial-state-with-using-es6-class-in-react).

This changeset actually have been done by an official React [migration tool](https://github.com/reactjs/react-codemod).

src/components/variable.jsx
fpagnoux: No, in React when you loop over a list you are supposed to provide a unique key for each element. Otherwise it throws a Warning.

PR #189, opened 2018-08-21
Reduce Bundle size

fpagnoux:
Depends on #187 

_Note: This PR is opened against `error-message` for diff readability, but should be merged on `master`. Target branch should be changed once #187  is merged._

Reduces the bundle size from 2.34MB to 1.47MB (-37%) by:
- Only bundling the used locales (fr and en)
- Only bundling the used syntax coloring (python)

1.47MB is still insanely high... The best way to reduce would be to get Swagger-ui out off the bundle and load in only when visiting the `/swagger` page

### DIFF
--- a/webpack.config.prod.babel.js
+++ b/webpack.config.prod.babel.js
@@ -69,6 +69,12 @@ module.exports = {
       new RegExp('^./(python)$')
     ),
 
+    // Only load locale for Fr and En
+    new webpack.ContextReplacementPlugin(
+      /react-intl\/locale-data$/,
+      new RegExp('^./(fr|en)$')
+    ),
+
     function() { this.plugin('done', writeAssets(path.resolve(__dirname, 'webpack-assets.json')).bind(this)) },
   ],
 }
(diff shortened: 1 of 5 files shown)


webpack.config.prod.babel.js
```diff
@@ -75,6 +75,18 @@ module.exports = {
...
+    // Only load syntax highlighting for Python
+    new webpack.ContextReplacementPlugin(
+      /highlight\.js\/lib\/languages$/,
+      new RegExp('^./(python)$')
+    ),
+
+    // Only load locale for Fr and En
+    new webpack.ContextReplacementPlugin(
+      /react-intl\/locale-data$/,
+      new RegExp('^./(fr|en)$')
```
sandcha: Set language according to configuration?

fpagnoux: Done!

webpack.config.prod.babel.js
```diff
@@ -75,6 +77,18 @@ module.exports = {
       },
     }),
 
+    // Only load syntax highlighting for Python
+    new webpack.ContextReplacementPlugin(
+      /highlight\.js\/lib\/languages$/,
+      new RegExp('^./(python)$')
+    ),
+
+    // Only load locale for Fr and En
```
sandcha: Update comment: `//Only load locale for supported languages`

PR #192, opened 2018-08-27
Add editorial content on variables and parameters 

### LINKED ISSUE #124: Add editorial content on a variable
As a variable maintainer, I can document useful information about a variable.
As a user of a tax and benefit information, I can know more precisely what is a variable, and how I can use it.

---

The legislation explorer extracts a lot of useful information from variables. However, we may want to keep and expose useful information that are not automatically extractible.

For instance:

- What are the typical input variables we should provide to compute a variable.
- For which period do they need to be specified
- Tricks about  the use of this variable ? (for instance, `loyer` is defined for a Menage but always used for a Famille. A line about how to easily cast it may help formula writers).

sandcha:
Fixes #124 

Add `Documentation:` field with multiline string value on variable and parameter pages.

On model's variable and parameter:
* Display `Documentation:` field when `documentation` attribute is returned by the web api. 
  * And keep new lines.
* Display nothing when `documentation` attribute doesn't exist in web api response.

Note: 
This PR ignores the following API attributes:
* parameter node `documentation` attribute
* variable formula `documentation` attribute

### DIFF
--- a/src/components/parameter.jsx
+++ b/src/components/parameter.jsx
@@ -31,10 +31,10 @@ class Parameter extends React.Component {
             : <em><FormattedMessage id="noDescription"/></em>
           }
           { parameter.documentation &&
-            (<span><FormattedMessage id="documentationText"/></span>)
-          }
-          { parameter.documentation &&
-            (<p className="documentation">{ parameter.documentation.trim() }</p>)
+            (<div>
+              <span><FormattedMessage id="documentationText"/></span>
+              <p className="documentation">{ parameter.documentation }</p>
+            </div>)
           }
           <div className="row">
             <div className="col-lg-8">
--- a/src/components/parameter.jsx
+++ b/src/components/parameter.jsx
@@ -30,6 +30,12 @@ class Parameter extends React.Component {
             ? <p className="description">{parameter.description}</p>
             : <em><FormattedMessage id="noDescription"/></em>
           }
+          { parameter.documentation &&
+            (<span><FormattedMessage id="documentationText"/></span>)
+          }
+          { parameter.documentation &&
+            (<p className="documentation">{ parameter.documentation.trim() }</p>)
+          }
           <div className="row">
             <div className="col-lg-8">
               {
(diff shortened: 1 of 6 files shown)


sandcha:
Parameter example:

screen shot 2018-08-27 at 23 27 44

Variable example:

screen shot 2018-08-27 at 23 28 55

src/components/parameter.jsx
```diff
@@ -30,6 +30,12 @@ class Parameter extends React.Component {
             ? <p className="description">{parameter.description}</p>
             : <em><FormattedMessage id="noDescription"/></em>
           }
+          { parameter.documentation &&
+            (<span><FormattedMessage id="documentationText"/></span>)
+          }
+          { parameter.documentation &&
```
fpagnoux: This block should be merged in the previous one, as the condition is the same.
You may need to wrap the `span` and `p` in a higher-level HTML element, as React don't like juxtaposing elements, but that should be doable 🙂

src/components/parameter.jsx
```diff
@@ -30,6 +30,12 @@ class Parameter extends React.Component {
             ? <p className="description">{parameter.description}</p>
             : <em><FormattedMessage id="noDescription"/></em>
           }
+          { parameter.documentation &&
+            (<span><FormattedMessage id="documentationText"/></span>)
```
fpagnoux: Are these parenthesis really useful?

src/components/parameter.jsx
```diff
@@ -30,6 +30,12 @@ class Parameter extends React.Component {
...
+          }
+          { parameter.documentation &&
+            (<p className="documentation">{ parameter.documentation.trim() }</p>)
```
fpagnoux: The trimming is not necessary if we [clean the values on the API side](https://github.com/openfisca/openfisca-core/pull/717/files#r219781241) :)

If the first use of this new field need trimming, this _should_ probably be done on the API side.

src/components/variable.jsx
```diff
@@ -174,6 +174,12 @@ class Variable extends React.Component {
               )}
           </ul>)
         }
+        { variable.documentation &&
```
fpagnoux: See comments on `parameters.jsx`

src/components/parameter.jsx
sandcha: Oops. Done.

src/components/parameter.jsx
sandcha: Fixed 😳

PR #200, opened 2018-10-02
Match pathname before :name in routes

### LINKED ISSUE #199: Style error on landing page first load
`.loading` CSS rule should not be applied to the list of parameters and variables:

capture d ecran 2018-10-02 a 18 24 16

## Reproduce it locally

Set your `nginx` config to something like this:

```
http {
  upstream legislation-explorer {
    server localhost:2030;
  }

  server {
    listen      80;
    server_name localhost;

    location ~/legislation(.*)$ {
        proxy_pass http://legislation-explorer$1;
        proxy_set_header Host $http_host;
        proxy_http_version 1.1;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }
  }
}
```

Then run:

```
npm install
npm run clean
npm run build
npm start
```

And visit `http://localhost/legislation`

bonjourmauko:
Fixes #199 

By matching pathname before :name in routes, we're sure it is exluded
from the :name match, hence not triggering the ParameterOrVariable
component rendering.

This is what was causing the loading div to be rendered in home page,
having as a side effect the style bug.

To test, follow #199 example.

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import {Route, IndexRoute} from 'react-router'
 
+import config from './config'
 import App from './components/app'
 import HomePage from './components/pages/home'
 import SwaggerPage from './components/pages/swagger'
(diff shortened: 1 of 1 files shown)


sandcha (approved):
🎉

PR #205, opened 2018-10-07
Use configured basename in server-side rendering

### LINKED ISSUE #202: Links on search results have different paths on server and client side rendering
Initial title: "Intermittent 404 on following links to variables pages"

(As observed & reproduced on 2 different machines during Openfisca demo 2018-10-03)

Morendil:
This *seems* to fix the problem, I'd feel better if we also had a unit test against this. On the other hand, this is a production bug so I'd prefer not to delay too long if we don't already have ways in place for checking this kind of property.

Fixes #202

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -17,7 +17,7 @@ function startServer(state) {
   server.use(favicon(path.resolve(__dirname, '../assets/favicon.ico')))
   server.use(express.static(path.resolve(__dirname, '../../public')))
 
-  server.use(handleRender(state))
+  server.use(handleRender(config.pathname, state))
 
   // Generic server errors (e.g. not caught by components)
   server.use((err, req, res, next) => {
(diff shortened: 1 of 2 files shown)


maukoquiroga:
Hi @Morendil, could you please add to the description de way you built, ran and tested the issue ?

> I'd feel better if we also had a unit test against this

I'm not sure how we could unit test this, seems more like an integration problem to me.

Morendil:
@maukoquiroga ISTM all the relevant info is in #202 - what seems needed from your point of view to get this to approved ?

I've taken a first look at the Watai tests, I think that's where an integration test might belong, but a) I'm not sure how to ensure it tests both client and server side renderings, and b) I'm going to need help understanding how to read and write Watai tests.

maukoquiroga:
I tested by hand because couldn't easily create a scenario with Watai.

I do not know if this solves the problem, but I'm confident this is a correct implementation.

I think we should take a breath and revisit the `history` hack.

PR #207, opened 2018-10-15
Fix date offset for visitors far from the UTC timezone

fpagnoux:
If I visit the legislation explorer at a time and from a timezone where the date is not the same one that in the UTC timezone, there is an offset in all the dates.

This is harder to see from France (except from 11pm to midnight), but probably more problematic for Aotearoa-NZ.

Before, at 6PM EST:

image

After, at 6PM EST:

image

### DIFF
--- a/src/components/formula.jsx
+++ b/src/components/formula.jsx
@@ -26,8 +26,8 @@ class Formula extends React.Component {
           <h3>
             <FormattedMessage id={stopDate ? 'fromToDate' : 'fromDate'}
               values={{
-                startDate: startDate && <FormattedDate value={startDate} year="numeric" month="2-digit" day="2-digit" />,
-                stopDate: stopDate && <FormattedDate value={stopDate} year="numeric" month="2-digit" day="2-digit" />,
+                startDate: startDate && <FormattedDate value={startDate} year="numeric" month="2-digit" day="2-digit" timeZone="UTC"/>,
+                stopDate: stopDate && <FormattedDate value={stopDate} year="numeric" month="2-digit" day="2-digit" timeZone="UTC"/>,
               }}
             />
           </h3>
--- a/src/components/formula.jsx
+++ b/src/components/formula.jsx
@@ -26,8 +26,8 @@ class Formula extends React.Component {
           <h3>
             <FormattedMessage id={stopDate ? 'fromToDate' : 'fromDate'}
               values={{
-                startDate: startDate && <FormattedDate value={startDate} year="numeric" month="2-digit" day="2-digit" timeZone="UTC"/>,
-                stopDate: stopDate && <FormattedDate value={stopDate} year="numeric" month="2-digit" day="2-digit" timeZone="UTC"/>,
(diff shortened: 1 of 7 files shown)


src/components/formula.jsx
```diff
@@ -26,8 +26,8 @@ class Formula extends React.Component {
           <h3>
             <FormattedMessage id={stopDate ? 'fromToDate' : 'fromDate'}
               values={{
-                startDate: startDate && <FormattedDate value={startDate} year="numeric" month="2-digit" day="2-digit" />,
-                stopDate: stopDate && <FormattedDate value={stopDate} year="numeric" month="2-digit" day="2-digit" />,
+                startDate: startDate && <FormattedDate value={startDate} year="numeric" month="2-digit" day="2-digit" timeZone="UTC"/>,
```
fpagnoux: > As no timezone is specified in `startDate`, ReactIntl assumes it's a UTC date. However, it will try to display it in the local timezone, hence the offset.
> Forcing the date formatting to UTC prevents this offset

Br3nda:
This leads me to the question: When we supply dates, we probably need to supply them in UTC - which is ambigious without a time attached.. We are currently 13 hours ahead of UTC. 

It's all a bit murky for date-of-birth, given we consider someone to be a year older once their birthday has occurred in New Zealand, even if they were born in France.

src/components/formula.jsx
Br3nda: We can't give dates in UTC. We'd have to ask the user if it happened before or after 1pm, which isn't very good UX. Generally the conditions like a parameter change, or a birthday is reached at midnight in the timezone, not midnight in UTC

Br3nda: Maybe the timezone is a configuration variable. We can configure that dates and times provided by the user are in NZ timezones. Most people won't understand the concept of providing for example their date-of-birth in UTC.

Morendil:
@Br3nda Oh… so a legal "date of birth" is not a DateTime in the programmer sense. That's probably listed somewhere [in there](https://github.com/kdeldycke/awesome-falsehood#dates-and-time). ;)

More seriously: we actually have a [related issue](https://github.com/openfisca/openfisca-france/issues/782) in the France model, where even *without* timezone offsets we had off-by-one errors in age calculations. I checked what concept of "age" the law uses and it matched your description above, you're considered a year older right after the clock ticks over on the date of your birth. So what would seem to matter is "local midnight". If a birth date input variable is specified without a time (which is what I'd expect), the conversion to a full datetime should do the right thing, i.e. convert both dates to datetimes at local midnight.

It's interesting that [your formulation of `age`](https://github.com/ServiceInnovationLab/openfisca-aotearoa/blob/6de1bc8f1401e7a63c1b0c356404cfefee114a08/openfisca_aotearoa/variables/demographics/ages.py#L28) looks quite different from [the one in France](https://github.com/openfisca/openfisca-france/blob/47e8dbbc7e07023237ee1ce582fb181a8bf57640/openfisca_france/model/prelevements_obligatoires/impot_revenu/ir.py#L59)…

PR #217, opened 2019-02-06
Fix text flow problem in country package info

### LINKED ISSUE #159: Country version text overflow problem in landing (mobile)
While navigating using a mobile phone,
I expect the country version text to be below the OpenFisca logo,
but instead it is shown as for the desktop layout,
which causes a text overflow problem thats unpleasant.

text overflow

Morendil:
Fixes #159 

Tested via SauceLabs "video" on Android emulated device. (This isn't ideal… I'm yearning for review apps.)

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -22,6 +22,10 @@ header .openfiscaPresentation p {
   margin-top: 1em;
 }
 
+#country-package-info {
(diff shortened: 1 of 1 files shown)


maukoquiroga:
capture d ecran 2019-02-07 a 22 50 14

PR #223, opened 2020-01-31
Fix integration testing

bonjourmauko:
Integration testing was failing because :

- Only `saucelabs` < 2 is supported by `watai`
- Only `sauceconnect` >= 4.4.12 is currently supported


Br3nda (approved):
Looks fixed. Lgtm

PR #235, opened 2022-12-13
Update History

bonjourmauko:


maukoquiroga:
@maukoquiroga or @sandcha or @cbenz Do you understand the error? Is it because incompatible dependencies of `watai`?

bonjourmauko:
@MattiSG  @sandcha @cbenz Do you understand the error? Is it because incompatible dependencies of `watai`?

